PR #3, opened 2015-06-09
Fix class name typo at TextRunMarshaller.php

stellalie:
This typo prevents marshall/unmarshalling functionality to work

### DIFF
--- a/src/qtism/data/storage/xml/marshalling/TextRunMarshaller.php
+++ b/src/qtism/data/storage/xml/marshalling/TextRunMarshaller.php
@@ -66,6 +66,6 @@ class TextRunMarshaller extends Marshaller
 	 */
     public function getExpectedQtiClassName()
     {
-        return 'randomInteger';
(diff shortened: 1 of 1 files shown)


bugalot:
Thanks a lot stellalie !

I'm just amazed that this did not appear in the unit tests before. Under which circumstances did you find this issue? I'm just curious :) !

All the best,
Jérôme

bugalot:
Moreover, I created a release 0.9.5 including your bugfix in the legacy branch. Please do not hesitate to update your composer.json.

If you have any trouble with this new release, just let me know!

Thanks again for the fix!

Bests,
Jérôme

stellalie:
@bugalot Thanks for that! That was very quick. It was a lucky catch actually when I was playing around with the repo. I was trying to convert a QTIComponent content to just plain xhtml string using MarshallerFactory so I can convert it to DOMElement then save that as XML string, and just happen to caught it :)

PR #11, opened 2015-08-27
Microseconds

Alroniks:
Partially related to this task https://oat-sa.atlassian.net/browse/TAO-1106

Added support of microseconds to durations

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
 	"name": "qtism/qtism",
 	"description": "OAT QTI Software Module Library",
 	"type": "library",
-	"version": "0.9.11",
+	"version": "0.9.12",
 	"authors": [
 		{
 			"name": "Open Assessment Technologies S.A.",
@@ -25,7 +25,8 @@
 		"GPL-2.0"
 	],
 	"require" : {
-		"php" : ">=5.3.0"
+		"php" : ">=5.3.0",
+	  	"league/period": "2.4.*"
 	},
 	"require-dev": {
 		"phpunit/PHPUnit": ">=3.7.20"
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
 		"GPL-2.0"
 	],
 	"require" : {
-		"php" : ">=5.3.0"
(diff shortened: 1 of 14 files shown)


Alroniks:
Ping @bugalot

bugalot:
Currently under review!

bugalot:
Hi @Alroniks !

Thanks for this PR.

It seems that the unit tests fail randomly. I spent an afternoon to understand what is going wrong, but without success. I can make appear errors at the level of Duration comparisons (equals method) when the CPU is under heavy load (I used cpuburn package with ubuntu). Comparison errors are very small, but exists. Usually, it breaks between a "non-microsecond" duration (e.g. PT5S) and another one with a very small difference (e.g. PT5.000001S). I assume that two DateTime objects are compared (diff method) but the second DateTime object has been instantiated some CPU cycles too late. This is why it appears when the CPU is under heavy load. It makes appear some kind of a "crital section" in the code that has to be found.

Thanks to have a look at this.

Jérôme

Alroniks:
Thank you for notice, @bugalot I have some ideas how fix it, so will send updates after.

Alroniks:
@bugalot Could you provide information about tests that failed. I tried run tests under cpuburn on virtualbox (I use homestead box) and all tests was passed.

bugalot:
Basically, common\datatypes\DurationTest.php can make the error happen, let's say 1 time each 10 attempts (it's a mean). 1 cpuburn (burnP5) is not enough, you need a big full charge with multiple cpuburn processes running, to get 100% CPU usage.

Alroniks:
Use 8 burns:
```
vagrant@homestead:~$ burnP5 & burnP5 & burnP5 & burnP5 & burnP5 & burnP5 & burnP5 & burnP5 &
[1] 15763
[2] 15764
[3] 15765
[4] 15766
[5] 15767
[6] 15768
[7] 15769
[8] 15770
```

And I call from 20 till 100 attemts for DurationTest.php and all passes for me. I will try run tests in mac (not in virtualbox), but don't know how catch it.

PR #12, opened 2015-09-04
Add callbacks to AssessmentItemSession.php

hutnikau:


bugalot:
Hi @hutnikau thanks for this PR!

Would you like please add some test cases?

All the best,
Jérôme

hutnikau:
@bugalot I have added unit test for this functionality.

bugalot:
Great! I think we are very close @hutnikau. Could you please put the runCallback method as a protected method (as far as I understand, the runCallback should never be called from the outside), and add in the registerCallback documentation what are the available event names, and what kind of parameters the callees will receive.

Thank you so much,
Jérôme

hutnikau:
@bugalot done!

bugalot:
Excellent @hutnikau. Everything looks fine when looking at the code. I will review this PR tomorrow morning. This is the first think I will do.

PR #13, opened 2015-09-21
Trial to solve moveNextAssessmentSection/TestPart if last item

bugalot:
Trial to solve moveNextAssessmentSection/TestPart if last item is timed out...


bugalot:
Oleg found the final issue in PR #14 !!!

PR #15, opened 2015-09-23
Fix last item timeout

hutnikau:


bugalot:
Awesome thx @hutnikau ! 

The issue with atan2 is quite fun. The error never appeared in the automated unit tests (Travis-CI). It's using Linux OS. Do you think we are facing a Windows only issue?

Bests,
Jérôme

hutnikau:
No, this is no due to windows. I asked a lot of people execute this command on different versions of php and on different OS. I got NAN on win10, win7 and Ubuntu. But also some of those whom I have asked got `-2.3561944901923` on win7. I have no idea what the reason of such behavior.
Win10:
nan
Win7:
nan2

bugalot:
Oleg, this is actually good information!

Indeed, it seems it is not related to Windows but maybe to particular PHP versions. I've seen in the past strange behaviours changing PHP from MINOR versions to MINOR versions. If you can get from the same people you asked their PHP version, we might find what's going on and maybe, report to PHP dev team.... :+1:

bugalot:
Does that resonate to you ?

hutnikau:
I was going to do it today. I want to gather more information to try to identify patterns.

bugalot:
Thanks so much @hutnikau !

hutnikau:
https://bugs.php.net/bug.php?id=70568

PR #20, opened 2015-11-11
Marshalling sequence changed, in order to be compatibale with xsd schema

boajer:
Pinging you, @bugalot , as can't assign directly


bugalot:
Could merge this one because it does not contain the invalid unit test from master.

PR #21, opened 2015-11-11
Marshalling sequence changed, in order to be compatibale with xsd schema

boajer:
Pinging you, @bugalot , as can't assign directly


bugalot:
Good catch @boajer !

However, as you can see in the Travis-CI report, the PR makes one unit test failing. Can you just adapt it so I can validate your PR?

Thanks again!

PR #22, opened 2015-11-18
Check select attribute value

hutnikau:


bugalot:
Excellent catch! Thanks a lot!

PR #23, opened 2015-11-20
AssessmentTestSession: fix condition to resume a session

jsconan:
Without this patch it is impossible to resume a suspended session. The `isRunning()` method considers the session is running when the state is different from INITIAL or CLOSED. So a suspended session is never considered as not running, and cannot be resumed.


bugalot:
Thank's I am currently reviewing!

PR #25, opened 2015-12-29
Change dependencies

llecaque:
rely on oatsa repo instead of Alroniks' one


boajer:
Hi @bugalot, do you think this PR can be merged?

PR #33, opened 2016-05-20
get new item collection instead of loop through route

hutnikau:


boajer:
@ssipasseuth Could you take a look on this PR please?

PR #35, opened 2016-06-09
A fix for a bug which caused Iterator does not iterate when…

stellalie:
… we only have one children component

@bugalot Can you please check this? I'm not sure whether this is the right fix, but I think the iterator should still work even if we only one child component, no? Thanks in advance.


bugalot:
Thanks a lot @stellalie  ! Excellent catch !

PR #57, opened 2016-09-07
Improved class detection compatibility with WC3 standard

boajer:
@bugalot @llecaque 
This changes prevents failing delivery compilation with not so nice, but still standard class names separated with more that one space.

### DIFF
--- a/qtism/common/utils/Format.php
+++ b/qtism/common/utils/Format.php
@@ -355,7 +355,7 @@ class Format {
 	 * @return boolean
 	 */
 	static public function isClass($string) {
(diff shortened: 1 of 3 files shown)


qtism/common/utils/Format.php
```diff
@@ -355,7 +355,7 @@ static public function isString256($string) {
 	 * @return boolean
 	 */
 	static public function isClass($string) {
-	    $pattern = "/^(?:[^\s]+?(?:\x20){0,1})+$/";
+        $pattern = "/^(?:[^\s]+?(?:\x20){0,})+$/";
```
ssipasseuth: The regex looks fine as such a case is still compatible with the qti xsd (xsd validation tested during import)

PR #58, opened 2016-09-07
Improved class detection compatibility with WC3 standard

boajer:
@bugalot @llecaque 
This changes prevents failing delivery compilation with not so nice, but still standard class names separated with more that one space.


llecaque:
@bugalot Merge and tag on legacy

PR #87, opened 2017-05-17
Feature/itemsession fully responded

bugalot:
The QTI specification declares that an itemSession is responded if

- an attempt has been begun
- at least one of the response variable have its value different than its default value

So litterally, the isResponded() method returns whether or not an itemSession is "partially" responded.

This PR enables client code to know wheter or not an itemSession is "fully" responded.

### DIFF
--- a/test/qtismtest/runtime/tests/AssessmentItemSessionTest.php
+++ b/test/qtismtest/runtime/tests/AssessmentItemSessionTest.php
@@ -661,4 +661,169 @@ class AssessmentItemSessionTest extends QtiSmAssessmentItemTestCase
         $this->assertTrue($itemSession->isResponded());
         $this->assertTrue($itemSession->isResponded(false));
     }
+    
+    public function testIsRespondedMultipleInteractions1()
+    {
+        $doc = new XmlDocument();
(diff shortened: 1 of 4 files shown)


test/qtismtest/runtime/tests/AssessmentItemSessionTest.php
```diff
@@ -653,5 +659,171 @@ public function testIsRespondedValueNullDefaultNotNull()
...
+                    'RESPONSEA',
+                    Cardinality::SINGLE,
+                    BaseType::IDENTIFIER,
+                    new QtiIdentifier('ChoiceA')
+                )
+            ])
+        );
+        
+        $this->assertTrue($itemSession->isResponded());
+        $this->assertTrue($itemSession->isResponded(false));
```
tverhoof: Suppose that after Attempt 4, the user will attempt the item again, and choose the default value as response (Choice A --> Choice C). In that case, isResponded(false) will return false. Is that the expected behavior ?

bugalot: Indeed young apprentice ! This is the expected behaviour. The qti specification considers an item is responded when the value differs from the default value. In your scenario, this is not the case so we are good

tverhoof: Ok I'll merge the PR then

PR #90, opened 2017-05-18
compare single mapping result to the lower and upper bound

ssipasseuth:
Fixed https://oat-sa.atlassian.net/browse/TAO-4329

### DIFF
--- a/qtism/runtime/expressions/MapResponseProcessor.php
+++ b/qtism/runtime/expressions/MapResponseProcessor.php
@@ -106,7 +106,7 @@ class MapResponseProcessor extends ExpressionProcessor
                         }
                     }
 
-                    if(is_null($result)){
+                    if (is_null($result)) {
                         // No relevant mapping found, return mapping default.
                         $result = $mapping->getDefaultValue();
                     }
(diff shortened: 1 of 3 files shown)


qtism/runtime/expressions/MapResponseProcessor.php
```diff
@@ -97,14 +98,25 @@ public function process()
                         }
                         
                         if ($val instanceof Comparable && $val->equals($mapKey) || $val === $mapKey) {
-                            return new QtiFloat($mapEntry->getMappedValue());
+                            $result = $mapEntry->getMappedValue();
                         } elseif ($variable->getBaseType() === BaseType::STRING && $val === null && $mapKey === '') {
-                            return new QtiFloat($mapEntry->getMappedValue());
+                            $result = $mapEntry->getMappedValue();
```
jsconan: Maybe should you break the loop after having found what you were searching for.
Or maybe are you expecting to find the last matching entry?

ssipasseuth:
Version updated

ssipasseuth:
applied suggested changes 👍

PR #97, opened 2017-07-13
fixed the individual item weight

ssipasseuth:
This fixes the calculation of the sum of outcome maximum processing.
With this fix, only the items having the same weight identifier used in the outcome rule will have its weight taken into account. Before the fix, the same weight is applied to all items. (code greatly inspired from TestVariablesProcessor which works well)
I have a test sample (the one with the weight) that can be used to test this behaviour, in the JIRA https://oat-sa.atlassian.net/browse/TAO-4687

### DIFF
--- a/qtism/runtime/expressions/OutcomeMaximumProcessor.php
+++ b/qtism/runtime/expressions/OutcomeMaximumProcessor.php
@@ -97,7 +97,7 @@ class OutcomeMaximumProcessor extends ItemSubsetProcessor
                     } else {
                         // If any of the items in the given subset have no declared maximum
                         // the result is NULL
-                        return NULL;
+                        return null;
                     }
                 } else {
                    // If any item has no SCORE variable, it may be an informational item with no response processing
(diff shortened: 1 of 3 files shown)


ssipasseuth:
please hold on this PR for now, I think I have found another issue

ssipasseuth:
I made the change to allow continuing the calculation of the outcomeMaximum when informational items or human based scoring items are found.
Ok ready for review now.

qtism/runtime/expressions/OutcomeMaximumProcessor.php
```diff
@@ -94,12 +95,13 @@ public function process()
                             $result[] = new QtiFloat(floatval($normalMaximum *= $weight->getValue()));
                         }
                     } else {
-                        // If any of the items in the given subset have no declared maximum
-                        // the result is NULL.
-                        return null;
+                        // If any of the items in the given subset have no declared maximum, possibly an item require human-based scoring)
+                        // we skip to the next item
+                        continue;
```
bugalot: I'm afraid we cannot accept this part of the PR, because of the following statement about the outcomeMaximum expression in the QTI Specification:

> If any of the items within the given subset have no declared maximum the result is NULL, otherwise the result has cardinality multiple and base-type float.

I think that a strategy must be found in the assessmentTest definition we generate to circumvent this. For instance, we could tag the items that have to be taken into account (non-information) using a category "non-informational" to be used with the outcomeMaximum.

```xml
<outcomeMaximum outcomeIdentifier="SCORE" includeCategory="non-informational" weight="SOMEWEIGHT"/>
```

PR #102, opened 2017-07-21
Preserver white spaces in XmlDocument

hutnikau:
During import of test package spaces in stimulus content which present between spans are omitted.
Before import:
```
<span>A</span> <span>B</span>
```
After import
```
<span>A</span><span>B</span>
```
Document component before and after fix:
space_2
space_1


bugalot:
Thank you very much for your contribution @hutnikau. Could you please make the same PR on the **develop** branch so that I can first test it on the QTI-SDK 8.6.X and then go ahead on the legacy version? Thanks so much in advance !

PR #113, opened 2017-09-07
Use charactermap for quick id validation

jbout:
As during serialization the identifiers get rechecked, there are millions of calls to this function, hence the regular expressions are becoming a performance bottleneck.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
 	"name": "qtism/qtism",
 	"description": "OAT QTI Software Module Library",
 	"type": "library",
-	"version": "0.12.2",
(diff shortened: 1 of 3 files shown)


jbout:
@bugalot Sorry for the ugly, but fast solution

bugalot:
@jbout no problem but I will be happy to know the explanation and how it could affect the validation :)

jbout:
The solution is basically a precalculated map of all allowed characters in the first or subsequent position, that we lookup character for character. Tested with random bytes against the previous implementation and should validate exactly the same strings as the previous implementation did.

PR #131, opened 2018-09-10
Minor Change To InterpolationTableEntry

CBrowning0791:
This fixes a thrown exception as sometimes the InterpolationTableEntryMarshaller will occasionally grab the $sourceValue as a double, and since a float and a double in PHP are the same thing.


bugalot:
Hello @CBrowning0791 !

As per PHP's `is_double` function documentation (http://php.net/manual/en/function.is-double.php), it is an alias of the `is_float` function. In such a context, I don't really see the point. Maybe I am missing something?

bugalot:
Closed because no activity since Oct 18, 2018

PR #134, opened 2019-04-24
for all `setOutcomeValue` elements in the response processing (templa…

zagovorichev:
…te or not) we need to check that the identifier value has a corresponding `outcomeDeclaration`. for that I need this rules.


zagovorichev:
rejected, changes need to be done in the legacy branch

PR #135, opened 2019-04-24
Small refactoring to get opportunity to get response processing rules

zagovorichev:


qtism/runtime/processing/ResponseProcessingEngine.php
```diff
@@ -146,51 +145,58 @@ protected function &getTemplateMapping() {
...
 			$this->trace(count($rules) . " responseRule(s) extracted from the response processing template");
 		}
-		
-		foreach ($rules as $rule) {
-			$engine = new RuleEngine($rule, $this->getContext());
-			$engine->process();
-			$this->trace($rule->getQtiClassName() . ' executed');
-		}
+
+		return (array) $rules;
```
bugalot: This is were the issue comes from @zagovorichev ! If we do not cast to array everything is fine.

A caller to the `getResponseProcessingRules()` can still get an array copy by calling `$rules->getArrayCopy()`

PR #139, opened 2019-07-31
Feature/title labels assessment item refs

bugalot:
This PR aims at bringing back `assessmentItem->title` and `assessmentItem->label` attribute values at the `assessmentItemRef` level while compacting `assessmentTest` documents.

### DIFF
--- a/src/qtism/data/storage/xml/schemes/qticompact_v1p0.xsd
+++ b/src/qtism/data/storage/xml/schemes/qticompact_v1p0.xsd
@@ -23,6 +23,8 @@
 	    			<xs:attribute name="adaptive" default="false" use="optional" type="xs:boolean"/>
 	    			<xs:attribute name="timeDependent" use="required" type="xs:boolean"/>
 	    			<xs:attribute name="endAttemptIdentifiers" use="optional" type="xs:string"/>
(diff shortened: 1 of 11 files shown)


src/qtism/data/storage/xml/schemes/qticompact_v1p0.xsd
```diff
@@ -23,6 +23,8 @@
 	    			<xs:attribute name="adaptive" default="false" use="optional" type="xs:boolean"/>
 	    			<xs:attribute name="timeDependent" use="required" type="xs:boolean"/>
 	    			<xs:attribute name="endAttemptIdentifiers" use="optional" type="xs:string"/>
+                    <xs:attribute name="title" use="optional" type="xs:string"/>
```
ekkinox: indentation issue

bugalot: Actually I have some indentation issues in this file. I would like to solve them after merging the PR so that we still see what is the exact modification. If not, the change will be overwhelmed between lot's of indentation fixes and will be difficult to track.

OK for you ?

ekkinox: deal

PR #140, opened 2019-08-06
Make some methods publicly available

bugalot:


src/qtism/runtime/tests/AssessmentTestSession.php
```diff
@@ -2182,11 +2182,13 @@ protected function addItemSession(AssessmentItemSession $session, $occurence = 0
...
      * Get an assessment item session.
      *
+     * Get an AssessmentItemSession object based on $assessmentItemRef and $occurence.
+     *
      * @param \qtism\data\AssessmentItemRef $assessmentItemRef
      * @param integer $occurence
      * @return \qtism\runtime\tests\AssessmentItemSession|false
      */
-    protected function getItemSession(AssessmentItemRef $assessmentItemRef, $occurence = 0)
+    public function getItemSession(AssessmentItemRef $assessmentItemRef, $occurence = 0)
```
mike-ionut-mihai-sandu-tao: can we declare the return type also and the type-hint for the occurrence 
also i think ```$occurence``` should be '$occurrence'

bugalot: Actually we cannot go for strict-typing for backward compatibility reasons, so we will have to rely on PHPDoc only for this ATM.

Regarding the typo with `$occurence` I'm sorting this out straight away ;)

PR #145, opened 2019-10-04
Fix/ordered container php72

bugalot:
This PR fixes a `warning` message `PHP Warning:  count(): Parameter must be an array or an object that implements Countable` in OrderedContainer with PHP >= 7.2

### DIFF
--- a/test/qtismtest/runtime/common/OrderedContainerTest.php
+++ b/test/qtismtest/runtime/common/OrderedContainerTest.php
@@ -12,12 +12,14 @@ use qtism\common\datatypes\QtiDirectedPair;
 use qtism\runtime\common\MultipleContainer;
 use qtism\common\enums\BaseType;
 
-class OrderedContainerTest extends QtiSmTestCase {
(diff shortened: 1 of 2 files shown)


test/qtismtest/runtime/common/OrderedContainerTest.php
```diff
@@ -12,32 +12,37 @@
...
-class OrderedContainerTest extends QtiSmTestCase {
+class OrderedContainerTest extends QtiSmTestCase
+{
 	
 	/**
 	 * @dataProvider equalsValidProvider
 	 */
-	public function testEqualsValid($containerA, $containerB) {
+	public function testEqualsValid($containerA, $containerB)
+    {
```
peetya: Wrong indent in L22 L31, L37 and L45. Maybe because of `space` vs `tab`

bugalot: Thanks @peetya ! solved

PR #148, opened 2019-10-14
Skpping should be allowed for any item having no resps. decl.

bugalot:
This PR aims at making items without interactions (a.k.a. informational items) always "skippable", as no "non-default" values can be provided as a response to such items.

### DIFF
--- a/src/qtism/runtime/tests/AssessmentItemSession.php
+++ b/src/qtism/runtime/tests/AssessmentItemSession.php
@@ -1361,7 +1361,7 @@ class AssessmentItemSession extends State
      */
     private function checkAllowSkipping(State $responses)
     {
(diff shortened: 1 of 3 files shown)


src/qtism/runtime/tests/AssessmentItemSession.php
```diff
@@ -1361,7 +1361,8 @@ public function templateProcessing()
      */
     private function checkAllowSkipping(State $responses)
     {
-        if ($this->getSubmissionMode() === SubmissionMode::INDIVIDUAL && $this->getItemSessionControl()->doesAllowSkipping() === false) {
+        // In case they are no response variable at all, the item is "skippable" as there is no possibility to provide an answer.
```
peetya: "In case **there** are ..."

PR #152, opened 2019-12-06
Add version to QTI Binary Storage.

bugalot:
This PR aims at providing the QTI Binary Storage format with a version number. This will enable us to extend the QTI Binary Storage in the future, if required.

### DIFF
--- a/src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
+++ b/src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
@@ -53,6 +53,8 @@ use \InvalidArgumentException;
  */
 abstract class AbstractQtiBinaryStorage extends AbstractStorage
 {
+    const VERSION = 1;
+
     private $seeker;
 
     /**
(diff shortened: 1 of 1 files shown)


src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
```diff
@@ -53,6 +53,8 @@
  */
 abstract class AbstractQtiBinaryStorage extends AbstractStorage
 {
+    const VERSION = 1;
```
julien-sebire: Shouldn't we apply https://semver.org/ here?

src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
```diff
@@ -234,6 +239,9 @@ public function retrieve($sessionId)
             $stream->open();
             $access = $this->createBinaryStreamAccess($stream);
 
+            // -- Consume version (not used yet).
```
julien-sebire: May we expect issues when retrieving sessions that were saved without version?

src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
bugalot: I would say it's overkill for a binary protocol

src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
bugalot: Yes, we will release this as a breaking change.

PR #158, opened 2019-12-23
Added unique sourceId assertion.

julien-sebire:
This PR is a follow-up to the https://github.com/oat-sa/qti-sdk/pull/157, adding a check for `sourceId` unicity.
If asked to add a new session identifier with an already existing `sourceId`, a `DuplicateSourceIdException` is thrown.

### DIFF
--- a/src/qtism/data/results/Context.php
+++ b/src/qtism/data/results/Context.php
@@ -158,11 +158,16 @@ class Context extends QtiComponent
      * @param string $sourceId
      * @param string $identifier
      * @return $this
+     * @throws DuplicateSourceIdException when an already existing sourceId is given.
      */
     public function addSessionIdentifier(string $sourceId, string $identifier): self
     {
         $identifier = Utils::normalizeString($identifier);
 
+        if ($this->hasSessionIdentifierWithSourceId($sourceId)) {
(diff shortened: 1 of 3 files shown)


src/qtism/data/results/Context.php
```diff
@@ -182,4 +187,25 @@ public function hasSessionIdentifiers(): bool
...
+
+    /**
+     * Check if the context has a session identifier with the given sourceId.
+     *
+     * @param $sourceId
+     * @return bool
+     */
+    private function hasSessionIdentifierWithSourceId($sourceId): bool
+    {
+        if (!$this->hasSessionIdentifiers()) {
```
jbout: not required as in this case the foreach loop would be skipped and we would still return false

test/qtismtest/data/results/ContextTest.php
```diff
@@ -33,21 +34,38 @@ public function testAddSessionIdentifier()
...
     }
+
+    public function testAddSessionIdentifierWithDuplicateSourceIdThrowsException()
+    {
+        $sourceId = 'a source id';
+        $identifier1 = 'id1';
+        $identifier2 = 'id2';
+
+        $subject = new Context();
+        $this->assertFalse($subject->hasSessionIdentifiers());
```
jbout: please add a unit test with non identical source id, but identical identifier (should pass)

PR #178, opened 2020-02-21
Refactor/container legacy

julien-sebire:
Moved Container from `runtime/common` to `common/collections` to follow change made in main branch in this commit: https://github.com/oat-sa/qti-sdk/commit/56031a0d794c301b28597722a4d07b804ed7c026
Added a `@deprecated` tag and message.

Known affected external usages are:
- https://github.com/oat-sa/extension-tao-result-export/blob/2d620626b1f2c04c75c6ac5fa1192c01e29183b4/test/AllBookletsExportUtilsTest.php
- https://github.com/oat-sa/extension-tao-nbce/blob/dca8e2906adb539d316785dac6d4a699366e5078/model/runner/QtiRunnerService.php
- https://github.com/oat-sa/extension-tao-act/blob/e050360802941ce6a0764ec4b686ba7b3dd97c88/model/Ods/V1/OdsTestPayload.php
- https://github.com/oat-sa/extension-tao-testqti/blob/923116287e80f0da7896c45d9bc1ab698dc5f55a/helpers/class.TestRunnerUtils.php

PRs for replacement of these usages will be submitted when the changes in this PR will be released.


qtism/common/collections/AbstractCollection.php
```diff
@@ -401,13 +400,4 @@ public function resetKeys()
         $newData = array_values($this->dataPlaceHolder);
         $this->setDataPlaceHolder($newData);
     }
-
-    public function __clone()
```
bugalot, line 405: This cannot be removed!

bugalot: Collections must remain clonable properly!

julien-sebire:
> Container represent the _runtime_ QTI container type which is used at runtime. Why moving it in common? It is supped to be used only in runtime.
Indeed, i was following the change made in the main branch in this commit a long time ago: https://github.com/oat-sa/qti-sdk/commit/56031a0d794c301b28597722a4d07b804ed7c026
But it doesn't really makes sense since every usage is still in qtism/runtime and no usage has been made outside of this repository. I'll move the Container into runtime/common in the main branch instead, in order to have things converging.

PR #180, opened 2020-03-17
Fix/udi 26 polygon intersection

wazelin:
- UDI-26 fix(coords): counting intersections between a horizontal line from a selected point and a polygon's edge between its first and last vertices
- UDI-26 chore: split returning workflows of `QtiCoords::inside()`
- UDI-26 chore: don't call `count` on each loop iteration in `QtuCoords::inside()`

### DIFF
--- a/src/qtism/common/datatypes/QtiCoords.php
+++ b/src/qtism/common/datatypes/QtiCoords.php
@@ -139,9 +139,9 @@ class QtiCoords extends IntegerCollection implements QtiDatatype
             }
 
             $intersects = 0;
-            for ($i = 1; $i < count($vertices); $i++) {
-                $vertex1 = $vertices[$i - 1];
-                $vertex2 = $vertices[$i];
+            for ($limit = count($vertices), $i = 0, $j = $limit - 1; $i < $limit; $i++) {
+                $vertex1 = $vertices[$i];
+                $vertex2 = $vertices[$j];
 
                 if ($vertex1[1] === $vertex2[1] && $vertex1[1] === $point->getY() && $point->getX() > min($vertex1[0], $vertex2[0]) && $point->getX() < max($vertex1[0], $vertex2[0])) {
                     // we are on a boundary.
@@ -161,6 +161,8 @@ class QtiCoords extends IntegerCollection implements QtiDatatype
                         $intersects++;
                     }
                 }
+
+                $j = $i;
             }
 
             // If we passed through an odd number of edges, we are in the polygon!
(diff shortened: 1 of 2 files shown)


src/qtism/common/datatypes/QtiCoords.php
```diff
@@ -120,52 +120,58 @@ public function inside(QtiPoint $point)
...
+            $vertex1 = $vertices[$i];
+            $vertex2 = $vertices[$j];
+
+            if ($vertex1[1] === $vertex2[1] && $vertex1[1] === $point->getY() && $point->getX() > min($vertex1[0], $vertex2[0]) && $point->getX() < max($vertex1[0], $vertex2[0])) {
+                // we are on a boundary.
+                return true;
             }
 
-            $intersects = 0;
-            for ($i = 1; $i < count($vertices); $i++) {
```
wazelin: The problem was here.
It didn't check the edge between $vertices[count($vertices) - 1] and `$vertices[0]`, that's it.

src/qtism/common/datatypes/QtiCoords.php
```diff
@@ -120,52 +120,58 @@ public function inside(QtiPoint $point)
...
-                $vertex[] = $this[$i]; //y
-
-                $vertices[] = $vertex;
+        }
+
+        // we consider it is a polygon.
+        // - Transform coordinates in vertices.
+        // -- Use of the "point in polygon" algorithm.
+        $vertices = [];
+        for ($i = 0, $limit = $this->count(); $i < $limit; $i++) {
```
julien-sebire: Ok for removing the `count()` from the loop evaluation. But please put it in another line, to keep the `for` readable:

```suggestion
        $limit = $this->count();
        for ($i = 0; $i < $limit; $i++) {
```

src/qtism/common/datatypes/QtiCoords.php
```diff
@@ -120,52 +120,58 @@ public function inside(QtiPoint $point)
...
+            $vertex = [];
+            $vertex[] = $this[$i]; //x
+            $i++;
+            $vertex[] = $this[$i]; //y
+
+            $vertices[] = $vertex;
+        }
+
+        $intersects = 0;
+        for ($limit = count($vertices), $i = 0, $j = $limit - 1; $i < $limit; $i++) {
```
julien-sebire: Nice trick to make it loop between first and last vertices. ;-)
Please just move the additional initialisations out of the for statement, it's difficult to read:
```suggestion
        $limit = count($vertices);
        $j = $limit - 1;
        for ($i = 0; $i < $limit; $i++) {
```

PR #184, opened 2020-05-20
Added QTI version change method to amend namespace of assessmentTest for both regular and compact test definitions.

julien-sebire:
This PR allows simple QTI version change.
A TestAssessment element can be amended with different QTI versions. As to now, only versions 2.1 and 2.2 are taken in account. This will allow us to compile tests to QTI 2.2 version.

### DIFF
--- a/qtism/data/storage/xml/XmlCompactDocument.php
+++ b/qtism/data/storage/xml/XmlCompactDocument.php
@@ -116,7 +116,7 @@ class XmlCompactDocument extends XmlDocument
      */
     public static function createFromXmlAssessmentTestDocument(XmlDocument $xmlAssessmentTestDocument, Resolver $itemResolver = null, Resolver $sectionResolver = null)
     {
-        $compactAssessmentTest = new XmlCompactDocument();
+        $compactAssessmentTest = new XmlCompactDocument($xmlAssessmentTestDocument->getVersion());
         $identifier = $xmlAssessmentTestDocument->getDocumentComponent()->getIdentifier();
         $title = $xmlAssessmentTestDocument->getDocumentComponent()->getTitle();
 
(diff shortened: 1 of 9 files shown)


qtism/data/storage/xml/XmlDocument.php
```diff
@@ -443,30 +443,43 @@ public function includeAssessmentSectionRefs($validate = false)
...
         $rootElement->setAttributeNS('http://www.w3.org/2001/XMLSchema-instance', 'xsi:schemaLocation', "http://www.imsglobal.org/xsd/imsqti_${qtiSuffix} ${xsdLocation}");
     }
 
+    /**
+     * Changes to Qti version of the root element.
+     *
+     * @param string $toVersion
+     * @return self
+     */
+    public function changeVersion(string $toVersion)
```
siwane, line 476: Missing return types

qtism/data/storage/xml/XmlDocument.php
```diff
@@ -443,30 +443,43 @@ public function includeAssessmentSectionRefs($validate = false)
...
     {
-        $qtiSuffix = 'v2p1';
-        $xsdLocation = 'http://www.imsglobal.org/xsd/qti/qtiv2p1/imsqti_v2p1.xsd';
         switch (trim($this->getVersion())) {
             case '2.0':
                 $qtiSuffix = 'v2p0';
                 $xsdLocation = 'http://www.imsglobal.org/xsd/imsqti_v2p0.xsd';
                 break;
 
+            case '2.2':
```
siwane, line 452: What do u think about taking under account `2.0.0` and `2.2.0` (if exists of course)

qtism/data/storage/xml/XmlCompactDocument.php
```diff
@@ -313,9 +313,26 @@ public function schemaValidate($filename = '')
      */
     public function decorateRootElement(DOMElement $rootElement)
     {
-        $rootElement->setAttribute('xmlns', "http://www.imsglobal.org/xsd/imsqti_v2p1");
+        $namespaces = [
```
siwane, line 316: What do u think about externalize this snippet to a Factory? Here u r relying of ``ttp://www.imsglobal.org/xsd/imsqti_%s` pattern, fingercrossed they will stick to it

qtism/data/storage/xml/XmlDocument.php
julien-sebire: On this matter, all we have are 2.0, 2.1 and 2.2 (3.0 coming). I'll clean it up and ensure that the version is validated before creating the object.

qtism/data/storage/xml/XmlDocument.php
julien-sebire: This will be done after the merge with master. Else, too much to do to keep both branches aligned.

qtism/data/storage/xml/XmlCompactDocument.php
julien-sebire: You're right, I'll change that, because it's too messy. I tried to make validation of the version, but the compact versions are just out of sync.
If you don't mind, Jérôme, I'll change qticompact 1.0 to 2.1 and 1.1 to 2.2, to align qticompact versions to qti versions, because validation is a nightmare.

julien-sebire:
Replaced by https://github.com/oat-sa/qti-sdk/pull/203

PR #187, opened 2020-05-28
Added getValue for all DataTypes.

julien-sebire:
This PR allows the usage of getValue method for all non-QtiScalar data types, using the implemented __toString method of each of these types.


bugalot:
@julien-sebire I think we should close thise one.

julien-sebire:
Replaced by those two PRs:
https://github.com/oat-sa/qti-sdk/pull/192
https://github.com/oat-sa/qti-sdk/pull/194

PR #190, opened 2020-06-02
Used more self explanatory version names for test session binary storage and added branch after version number.

julien-sebire:
This PR sets up a versioning for test session binary storage compatible with master branch, so that we can discriminate between master and legacy logic in retrieving test session.
A letter `M` for master and `L` for legacy is added after the version number, starting from version 10.
All versions below 10 are considered legacy version.
Also, detection of version by feature name is more readable.


julien-sebire:
Thanks for the comment @bugalot !
It's been tested with the following scenarios using Tao demo "Qti Example Test" (9 items):
- test completely run with Tao Community sprint 130 (test session storage version 9): **OK** 
- test started with version 9 above, answered three items, switched to version 10 for other items: **OK**
- test completely run with version 10: **OK**
- test started with version 10, answered three items, switched to previous version 9 for next items: crashed on item submission = expected behavior: **OK**
- test started with version 9 above, answered three items, switched to version 10 for next two items, switched to version 9, crashed (expected), switched to version 10, worked well again, finished the test with version 10: **OK**

Conclusion:
- Execution of whole test in version 9 works: **OK**
- Execution of whole test in version 10 works: **OK**
- Transition from version 9 to 10 during test works: **OK**
- Transition from version 10 to 9 crashes: **OK**

It seems to me that this ensures that the transition to this new version is working correctly and can be merged.

Crash message when reverting from version 10 to 9:
tao-content-your-administrator

PR #192, opened 2020-06-14
Fix/nex 1091/value serialization

bugalot:
This should fix the issue raised by @peetya. I added additional unit tests @julien-sebire .

PS: I don't know why the Travis builds are not linked to this PR (maybe a Travis issue) but the build was successful over here -> https://travis-ci.org/github/oat-sa/qti-sdk/builds/698172922

### DIFF
--- a/src/qtism/runtime/common/Variable.php
+++ b/src/qtism/runtime/common/Variable.php
@@ -26,6 +26,7 @@ namespace qtism\runtime\common;
 use InvalidArgumentException;
 use qtism\common\collections\Container;
 use qtism\common\datatypes\QtiDatatype;
+use qtism\common\datatypes\QtiScalar;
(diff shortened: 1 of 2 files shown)


julien-sebire:
Hello Jérôme,
We went on the same path, not the same way (second PR for the same fix, sorry I didn't know you would work on this so I did it on my side: https://github.com/oat-sa/qti-sdk/pull/194)
I figured it out that the dataModel value for non-scalar values should return the actual Datatype object, not a value converted to string (for example, a QtiPair should be returned instead of a string representation of the pair).
It would have been perfect to have explicit tests about this at the moment of the first conception. We are all perfectible beings... ;-)

julien-sebire:
Duplicated with https://github.com/oat-sa/qti-sdk/pull/194.

PR #197, opened 2020-07-01
Feature TCA-666 aria-* attributes

bugalot:
This Pull Request aims at providing [`aria-*` attributes support in QTI](http://www.imsglobal.org/question/qtiv2p2/imsqti_v2p2_impl.html#7.1.2)-SDK `master`. This should result in a `13.1.0` release and happens in the context of [TCA-666](https://oat-sa.atlassian.net/browse/TCA-666) where there is an absolute need for aria-* attributes support. In such a context, we first apply this to the `master` version (NG), then we will tackle the `legacy` one (CG).

This touches multiple aspects of the QTI-SDK:

* Data Model: `aria-*` attributes are taken into account in the context of QTI 2.2.X. Out of this context, `aria-*` attributes are ignored to remain compliant with XML Schemas. See `qtism\data` namespace.
* Rendering: `aria-*` attributes are being rendered. See `qtism\runtime\rendering` namespace.

Here is my analysis on the required coverage of `aria-*` attributes based on crawling the [QTI 2.2.2 XSD](http://www.imsglobal.org/xsd/qti/qtiv2p2/imsqti_v2p2p2.xsd). This enabled me to know where to start implementing this within de Data Model. I did not have the time to check the exact similarity of [QTI 2.2.0](http://www.imsglobal.org/xsd/qti/qtiv2p2/imsqti_v2p2.xsd) and [QTI 2.2.1](http://www.imsglobal.org/xsd/qti/qtiv2p2/imsqti_v2p2p1.xsd). **I count on you guys for this.**

**Please pay attention to `aria-flowsto` vs `aria-flowto` issue in the specification.** This is tricky. @julien-sebire please put this on the table on a next QTI Task Force Meeting. This is very important. I put the necessary efforts to remain compliant in what we marshall/unmarshall in XML, but also put in place some fallbacks to make sure everybody is happy and that we cover the most of the existing userland (millions of QTI Assessment Materials of the outside world).

```
QTI 2.2.2 XSD (92)
------------------------------------------------------------------------

BasePromptInteraction.type (controls, describedby, flowsto, label, labelledby, level, live, orientation, owns): 14
---------------------------
associateInteraction -> BlockInteraction -> Interaction -> BodyElement
choiceInteraction -> BlockInteraction -> Interaction -> BodyElement
drawingInteraction -> BlockInteraction -> Interaction -> BodyElement
extendedTextInteraction -> BlockInteraction -> Interaction -> BodyElement
[...]

PR #198, opened 2020-07-02
Backport aria-* from master

bugalot:
This is the backport of the [`aria-*` attributes feature](https://github.com/oat-sa/qti-sdk/pull/197) performed on `master` version. Please have a deep look at it to understand the scope of the PR. This should result in a release `0.21.0` of QTI-SDK `legacy`.

Compared to the `master` approach, the major obstacle here is that we did not have yet "QTI Version Management" inside `Marshaller` components was not included yet. I had to backport it within the scope of this `aria-*` feature. That is why we have so much changes.

**Please be aware that that such a PR cannot be merged without a FULL INTEGRATION TESTING in TAO Current Gen**. With the millions of Assessment Materials around the world, and all the Assessment Materials that our customers have to deal with. Lots of changes were performed. **Failure is not an option as it could have catastrophic impacts on all our customer base**. Another round of QA like https://oat-sa.atlassian.net/browse/QI-316 must be performed, in addition with specific `aria-*` contents ingestion/delivery to test takers to make sure everything is fine.

As a positive and motivational outcome, this is the road to a better integration of QTI 2.2 in TAO Current Gen. With QTI version support in QTI-SDK `legacy`, we can really tackle essential QTI 2.2 support like `dir`, `bidi`, `data-*` attributes. There is a big added value here !

**TAO Integration Testing branches ares the following. Please take extra care with reading carefully the instructions in these 2 Pull Requests.**

* https://github.com/oat-sa/lib-tao-qti/pull/72
* https://github.com/oat-sa/extension-tao-itemqti/pull/1483

### DIFF
--- a/qtism/data/storage/xml/XmlCompactDocument.php
+++ b/qtism/data/storage/xml/XmlCompactDocument.php
@@ -56,6 +56,24 @@ use SplObjectStorage;
 class XmlCompactDocument extends XmlDocument
 {
     /**
+     * XmlCompactDocument constructor.
+     *
+     * Create a new XmlCompactDocument object.
+     *
+     * @param string $version
+     * @param QtiComponent|null $documentComponent
+     */
+    public function __construct($version = '2.1', QtiComponent $documentComponent = null)
+    {
+        // Version 1.0 was used in legacy code, let's keep it BC.
+        if ($version === '1.0') {
+            $version = '2.1.0';
+        }
+
+        parent::__construct($version, $documentComponent);
+    }
+
+    /**
      * Whether or not the rubricBlock elements
      * must be separated from the core document.
      *
--- a/qtism/data/storage/xml/XmlCompactDocument.php
+++ b/qtism/data/storage/xml/XmlCompactDocument.php
@@ -380,4 +380,9 @@ class XmlCompactDocument extends XmlDocument
 
         return $references;
     }
+
+    protected function inferVersion()
+    {
+        return '2.1';
+    }
 }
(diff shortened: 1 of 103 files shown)


qtism/data/storage/xml/XmlCompactDocument.php
```diff
@@ -55,6 +55,24 @@
...
+     * Create a new XmlCompactDocument object.
+     *
+     * @param string $version
+     * @param QtiComponent|null $documentComponent
+     */
+    public function __construct($version = '2.1', QtiComponent $documentComponent = null)
+    {
+        // Version 1.0 was used in legacy code, let's keep it BC.
+        if ($version === '1.0') {
+            $version = '2.1.0';
```
gitromba, line 70: image

bugalot: Unfortunately required for legacy.

julien-sebire: This won't be enough to get everything aligned for legacy and master, but it will be part of another PR.

qtism/data/storage/xml/XmlCompactDocument.php
```diff
@@ -380,4 +398,9 @@ public function explodeRubricBlocks()
 
         return $references;
     }
+
+    protected function inferVersion()
+    {
+        return '2.1';
```
julien-sebire, line 404: This will also need to be changed to enable compilation to QTI 2.2 tests. Will be part of another PR.

PR #201, opened 2020-07-28
Feature/add compilation version master 2

julien-sebire:
Up to now, Tao compiles tests to QTI 2.1 compact test.
The aim of this trail of PRs is to allow the compilation of any 2.x.x QTI test into a QTI 2.2 compact test, to include all the QTI 2.2 features. 

This PR adds the actual changeVersion feature, allowing to transform a QTI 2.x.x xml content into another 2.x.x version.
It works on QTI document, QTI compact documents and QTI result reports.

For an easier review, please follow the commits instead of the global change.
When reviewing, keep in mind that this PR is the minimal change to make it work. There's another PR following that takes care of refactoring and code-styling:
https://github.com/oat-sa/qti-sdk/pull/202

A regression test has been successfully carried out on VDS using custom branches:
https://oat-sa.atlassian.net/browse/MS-152?focusedCommentId=93640


src/qtism/data/storage/xml/XmlResultDocument.php
```diff
@@ -38,7 +40,30 @@ class XmlResultDocument extends XmlDocument
...
+            throw new InvalidArgumentException(
+                sprintf(
+                    'QTI Result Report is not supported for version "%s". Supported versions are "%s".',
+                    $versionNumber,
+                    implode('", "', $knownVersions)
+                )
+            );
+        }
+
+        return $filename;
```
epoxxid, line 66: Also, for me switch with immediate return, throwing exception at the bottom, would fit better here

src/qtism/data/storage/xml/XmlResultDocument.php
```diff
@@ -73,4 +98,28 @@ protected function decorateRootElement(DOMElement $rootElement)
         $rootElement->setAttributeNS('http://www.w3.org/2000/xmlns/', 'xmlns:xsi', 'http://www.w3.org/2001/XMLSchema-instance');
         $rootElement->setAttributeNS('http://www.w3.org/2001/XMLSchema-instance', 'xsi:schemaLocation', "http://www.imsglobal.org/xsd/imsqti_${qtiSuffix} ${xsdLocation}");
     }
+
+    protected function inferVersion()
```
epoxxid, line 102: Can we have a return type hint here?

src/qtism/data/storage/xml/XmlResultDocument.php
```diff
@@ -73,4 +98,28 @@ protected function decorateRootElement(DOMElement $rootElement)
...
+        $version = false;
+
+        if (empty($root) === false) {
+            $rootNs = $root->namespaceURI;
+
+            if ($rootNs === 'http://www.imsglobal.org/xsd/imsqti_result_v2p1') {
+                $version = '2.1.0';
+            } elseif ($rootNs === 'http://www.imsglobal.org/xsd/imsqti_result_v2p2') {
+                $version = '2.2.0';
+            }
```
epoxxid, line 115: Again, IMO doing a return immediately and throwing exception at the most bottom would improve readability

src/qtism/data/storage/xml/XmlDocument.php
```diff
@@ -729,4 +730,30 @@ protected function inferVersion()
...
+            $filename = __DIR__ . '/schemes/qtiv2p1p1/imsqti_v2p1p1.xsd';
+        } elseif ($version === '2.2.0') {
+            $filename = __DIR__ . '/schemes/qtiv2p2/imsqti_v2p2.xsd';
+        } elseif ($version === '2.2.1') {
+            $filename = __DIR__ . '/schemes/qtiv2p2p1/imsqti_v2p2p1.xsd';
+        } elseif ($version === '2.2.2') {
+            $filename = __DIR__ . '/schemes/qtiv2p2p2/imsqti_v2p2p2.xsd';
+        } else {
+            $filename = __DIR__ . '/schemes/imsqti_v2p0.xsd';
+        }
```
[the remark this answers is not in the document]
julien-sebire: This is done in the next PR.
Also, return types will be set globally on a codestyle-only PR later, to also help the progressive convergence of master and legacy branch.

src/qtism/data/storage/xml/XmlResultDocument.php
julien-sebire: Totally agree. In the next PR, you'll see that. Even more, <spoiler>we finally change for an object support that let's each version to just provide the right filename</spoiler>

src/qtism/data/storage/xml/XmlResultDocument.php
julien-sebire: Done in the next PR again... :)

src/qtism/data/storage/xml/XmlResultDocument.php
julien-sebire: Done in next PR... :)

src/qtism/data/storage/xml/XmlResultDocument.php
Haru0: I'm not convinced by putting a switch anywhere. Both are against the open-close principle. Perhaps this could be extracted to some kind of resolver?

julien-sebire: This is done in the next PR. We have a mechanism to identify which version the document is written for, and we then use a object representing this version, that provides all the filenames, xsd locations and all.

PR #205, opened 2020-08-12
Added assertions to check that the testsession is closed on each test.

julien-sebire:
Complementary PR to the previous one to have assertions for every case.


julien-sebire:
Completed tests on session ending when going to next testPart on the last testPart.

PR #221, opened 2020-09-10
Allowed 0 as an expectedLength value.

julien-sebire:
This PR aims at fixing the implementation of expectedLength for (Extended)TextEntryInteraction.
It allows 0 as a valid length and replace the usage of -1 as a magical value representing no value at all with a null value.

Related ticket:
https://oat-sa.atlassian.net/browse/NEX-1852


peetya (approved):
LGTM, nice work!

PR #163, opened 2020-01-15
Code style - no code change

julien-sebire:
The idea of this PR is to standardize both current and legacy branches to merge them in a not so far future:
- import and reduce FQCN
- remove extra @author tags not in the file docblock
- change secondary copyright date to 2020 for all changed files
- complete the methods doc blocks with missing tags (@param, @return, @throws, ...)
- array() to []
- double quotes to single quotes when applicable

All tests are run without any problem during the code styling process.

Related PR for legacy branch: https://github.com/oat-sa/qti-sdk/pull/164

PR #232, opened 2020-09-21
The Big Refactor In The Sdky - Part III - Main branch version

julien-sebire:
This is the third big bunch of refactoring in order to make both main and legacy branches to converge in a not so far future.
Part I: https://github.com/oat-sa/qti-sdk/pull/163
Part II: https://github.com/oat-sa/qti-sdk/pull/188

It includes:
- Fixed typos in `occurred`, `occurrence` and `whether` in strings and comments (not variable nor method names).
- `"` to `'` string delimiters when possible.
- Incremental operators `++`, `--`, `+=`, `-=` when possible.
- `self` keyword instead of class name.
- Removed useless class aliases in `use` statements.
- Removed useless local variables.
- Type casting instead of `*val` methods, e.g. `(int)$var` instead of `intval($var)`.
- `::class` and `use` statement instead of FQN.
- Null coalescing operator when possible.
- Updated PhpDocs for `@param`, `@result`, `@throws`.
- Replaced deprecated `setExpectedException` with `expectException*`.
- Removed `@see` tags referring to the parent method and useless `@inheritdoc` tags.
- Fixed wrongly static calls to dynamic calls when in an object occurrence context.
- Removed `=== true` for standard functions returning a boolean.
- Removed auto-generated `@package` tags. 
- Moved `isPure` to `Expression` (parent class).
- Fixed `setRule` in `RuleProcessor`.
- Moved `Container` back from `common/collections` to `runtime/common` as in legacy branch.
- Removed static calls to QtiSmTestCase::createDOMElement.
- And a lot of code cleaning...

Legacy branch counterpart: https://github.com/oat-sa/qti-sdk/pull/233


src/qtism/cli/Render.php
```diff
@@ -351,7 +346,7 @@ private function runXhtml(XmlDocument $doc, XhtmlRenderingEngine $renderer)
             $header .= "${indent}<head>${nl}";
             $header .= "${indent}${indent}<meta charset=\"utf-8\">${nl}";
 
-            if (empty($title) !== false) {
+            if (empty($title)) {
```
gitromba, line 349: You may want to add the title when it's **not** empty.
```suggestion
if (!empty($title)) {
```

julien-sebire: You have a point. This makes sense. Indeed, this is inverting the existing logic (`!== false` is equivalent to `=== true`), which seemed to be faulty. Fortunately, this is some code used only to demo the possibilities in CLI, that is not used in production and that is not covered by tests. I'll changed that.
Thanks for this Misi! ;-)

PR #227, opened 2020-09-14
Feature/nex 1877/uploaded file not in test session

julien-sebire:
This PR aims at allowing the storage of a hash of an uploaded file instead of the content of the file itself, to avoid huge payloads to be stored in the test session. This requires that the file has been **persisted to an external storage** prior to call to `qtism\runtime\pci\json\Unmarshaller::unmarshall`.
A convenience to trigger this process is to use the `FileHash::FILE_HASH_KEY` constant as a key when encoding the JSON payload. An **extra 4th parameter `id`** is then required, containing the id of the persisted file allowing the retrieval of the file by the application that persisted the file on the external storage:

Replace key `file` in
```
{ "base" : { "file" : {
            "mime" : "text-plain", 
            "data" : "The contents of the file...", 
            "name" : "my-file.txt" } } }
```
with key `FileHash::FILE_HASH_KEY`:
```
{ "base" : { "fileHash" : {
            "mime" : "text-plain", 
            "data" : "AA025C4DB95A39A2CB8525F83F1387C1A2B13595C8E4C337CDF9AD7B043518A3", 
            "name" : "my-file.txt",
            "id" : "some-identifier/provided/by-the/external/file-store" } } }
```

PR #234, opened 2020-10-27
Revert "Feature/nex 1877/uploaded file not in test session"

julien-sebire:
Reverts oat-sa/qti-sdk#227 Missing version management.


julien-sebire:
No need for that. All session persist is always to the last version and all read will be from previous version only once, so no overkill strategy pattern is really needed for one read only.

PR #240, opened 2020-12-28
test

julien-sebire:


julien-sebire:
Was just to test travis configuration for php >= 7.3.

PR #259, opened 2021-04-01
fix: do not escape starting and ending symbols for patterns validation

m-lagarde:
[TR-825](https://oat-sa.atlassian.net/browse/TR-825)

The goal of this PR is to fix the pattern control from validation constraint when starting and ending symbols are provided

### DIFF
--- a/src/qtism/runtime/expressions/operators/Utils.php
+++ b/src/qtism/runtime/expressions/operators/Utils.php
@@ -328,6 +328,8 @@ class Utils
      */
     public static function prepareXsdPatternForPcre($pattern)
     {
+        $pattern = trim($pattern, '^$');
+
         // XML schema always implicitly anchors the entire regular expression
         // because there is no carret (^) nor dollar ($) signs.
         // see http://www.regular-expressions.info/xml.html
--- a/src/qtism/runtime/expressions/operators/Utils.php
+++ b/src/qtism/runtime/expressions/operators/Utils.php
@@ -328,7 +328,7 @@ class Utils
      */
     public static function prepareXsdPatternForPcre($pattern)
     {
-        $pattern = trim($pattern, '^$');
+        $pattern = rtrim(ltrim($pattern, '^'), '$');
 
         // XML schema always implicitly anchors the entire regular expression
         // because there is no carret (^) nor dollar ($) signs.
--- a/test/qtismtest/runtime/expressions/operators/PatternMatchProcessorTest.php
+++ b/test/qtismtest/runtime/expressions/operators/PatternMatchProcessorTest.php
@@ -110,7 +110,9 @@ class PatternMatchProcessorTest extends QtiSmTestCase
             [new QtiString('string'), 'shell', false],
             [new QtiString('stringString'), '.*', true], // in xml schema 2, dot matches white-spaces
             [new QtiString('^String$'), 'String', false], // No carret nor dollar in xml schema 2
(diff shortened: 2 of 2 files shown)


src/qtism/runtime/expressions/operators/Utils.php
```diff
@@ -328,6 +328,8 @@ public static function lastPregErrorMessage()
      */
     public static function prepareXsdPatternForPcre($pattern)
     {
+        $pattern = trim($pattern, '^$');
+
         // XML schema always implicitly anchors the entire regular expression
         // because there is no carret (^) nor dollar ($) signs.
```
julien-sebire, line 334: This should read:
```
// XML schema always implicitly anchors the entire regular expression.
// Neither caret (^) nor dollar ($) sign are supported as anchors.
```
It's more explicit. You can change that in the same time.

test/qtismtest/runtime/expressions/operators/PatternMatchProcessorTest.php
```diff
@@ -110,7 +110,9 @@ public function patternMatchProvider()
             [new QtiString('string'), 'shell', false],
             [new QtiString('stringString'), '.*', true], // in xml schema 2, dot matches white-spaces
```
julien-sebire, line 111: This test doesn't actually test what the comment says. The searched string should contain at least one space:
```
[new QtiString('string String'), '.*', true], // in xml schema 2, dot matches white-spaces
```

test/qtismtest/runtime/expressions/operators/PatternMatchProcessorTest.php
```diff
@@ -110,7 +110,9 @@ public function patternMatchProvider()
...
             [new QtiString('^String$'), 'String', false], // No carret nor dollar in xml schema 2
-            [new QtiString('^String$'), '^String$', true],
+            [new QtiString('String'), 'String', true],
+            [new QtiString('String'), '^String$', true],
+            [new QtiString('^String$'), '^String$', false],
             [new QtiString('Str/ing'), 'Str/ing', true],
```
julien-sebire, line 116: Please add tests to match `12.99$` and `$12.99`.

julien-sebire (changes requested):
Nice try but, how can you match `10.99$` or `$10.99` then? :-) I'm not sure we could find use cases for matching `^` at beginning or end of a string though.
Maybe remove ^ and $ together only when they are both here?
Or give a hint in your front-end to tell that nor ^ nor $ anchors are supported and that the pattern will be matched to the whole response?
Or take them in account when escaped (`\^` and `\$`)?

src/qtism/runtime/expressions/operators/Utils.php
```diff
@@ -328,6 +328,8 @@ public static function lastPregErrorMessage()
      */
     public static function prepareXsdPatternForPcre($pattern)
     {
+        $pattern = trim($pattern, '^$');
```
wazelin, line 331:
```suggestion
        $pattern = rtrim(ltrim($pattern, '^'), '$');
```

wazelin (changes requested):

test/qtismtest/runtime/expressions/operators/PatternMatchProcessorTest.php
wazelin: Considering $ is usually prepended to a value, I would add just `$12.99` support.

PR #266, opened 2021-04-13
Feature/nex 2008/html5 main

julien-sebire:
Related ticket: https://oat-sa.atlassian.net/browse/TRN-75

This PR aims at adding support for the HTML5 tag `<track>` on *MediaInteraction*.
Since `<track>` tag is a sub-element of `<video>` and `<audio>` HTML5 tags, this implies the need to also support the following features:
- `<video>` and all its sub elements (`<source>`, `<track>`)
- `<audio>` and all its sub elements (`<source>`, `<track>`)
- namespaces in the serializing/unserializing system (marshallers)
- html5 generic elements
- usage of exclusively one among `<object>`, `<audio>` and `<video>` in *MediaInteraction*
- usage of `<audio>` and `<video>` in `ItemBody`
- HTML code generation (rendering)

Since HTML5 is an addition to QTI, all the HTML5 elements tags have to be prefixed with `qh5:` prefix as shown in this example: https://github.com/oat-sa/qti-sdk/pull/266/files#diff-bcc3940df1a5e18e0f7d4628d30dc35b3f5e4da66b7e2cf25e7283c428ea9e6f


julien-sebire:
Close as not necessary anymore, with html5 work done in separate PRs.

PR #270, opened 2021-04-26
Support for % sign in width and height attributes of <img> tag.

julien-sebire:
Continuation of #269 extended to  element.

### DIFF
--- a/qtism/data/content/xhtml/Img.php
+++ b/qtism/data/content/xhtml/Img.php
@@ -188,11 +188,9 @@ class Img extends AtomicInline
     }
 
     /**
-     * Set the height of the image.
+     * Set the height of the image. A null value unsets height.
      *
-     * A null value means that no height is provided.
-     *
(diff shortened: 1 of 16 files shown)


qtism/data/content/xhtml/Img.php
```diff
@@ -194,79 +188,62 @@ public function hasLongdesc()
...
-     * Whether a height attribute is defined.
-     *
-     * @return bool
+     * Whether a height is set.
      */
-    public function hasHeight()
+    public function hasHeight(): bool
     {
-        return $this->getHeight() >= 0;
+        return $this->height !== null;
```
pnal, line 217: Could it affect compiled deliveries with -1 set as default? 
e.g. it will return true but actually nothing been set

julien-sebire: Good catch! When we set -1, the value is set to null.

pnal (approved):
Good job!

PR #271, opened 2021-04-27
Support for % sign in width and height attributes of <img> and <object> tag - Main branch

julien-sebire:
This PR fixes support for `%` sign in height and width attributes of `<object>` and `` QTI/HTML elements.

### DIFF
--- a/src/qtism/data/content/xhtml/ObjectElement.php
+++ b/src/qtism/data/content/xhtml/ObjectElement.php
@@ -68,20 +68,28 @@ class ObjectElement extends BodyElement implements FlowStatic, InlineStatic
     private $type;
 
     /**
-     * The width. -1 means no height was provided.
+     * The object's width attribute.
      *
-     * @var int
+     * The value of this attribute can be:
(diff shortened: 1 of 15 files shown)


src/qtism/data/content/xhtml/ObjectElement.php
```diff
@@ -158,79 +164,62 @@ public function getType()
...
+        $this->width = Format::sanitizeXhtmlLength($width, 'width');
     }
 
     /**
      * Get the width of the object.
-     *
-     * A negative value describes that no width is provided.
-     *
-     * @return int A width.
+     * Null means that no width is set.
```
pnal, line 179: I think it is unnecessary or confusing comment, now return type says everything, and cause the class contains "has" methods it would be correct to say that their responsibility to answer whether width been set.
(Same for similar getters)

pnal (approved):
Good job!

src/qtism/data/content/xhtml/ObjectElement.php
julien-sebire: Thanks for remark. Done!

PR #274, opened 2021-05-07
Fix TR-1084: skip unnecessary variable instantiation

wazelin:
- feat: introduce `VariableFactoryInterface` and its implementation, allowing to instantiate any `Variable` from its `VariableDeclaration
- fix: skip redundant `Variable` instantiation and initialization upon `QtiBinaryStreamAccess::readAssessmentItemSession()` execution, and read the variable directly from the session state

This is not a _major_ improvement, but those variable's objects instantiation and initialization does take some time indeed, and I cannot call that time insignificant (around 10% of the `move` action request processing).
Perhaps, further improvement is possible, involving lazy initialization of the variable's internals.

Legacy release backport is #276.

### DIFF
--- a/src/qtism/runtime/common/VariableFactory.php
+++ b/src/qtism/runtime/common/VariableFactory.php
@@ -44,7 +44,9 @@ final class VariableFactory implements VariableFactoryInterface
      */
     public function createFromDataModel(VariableDeclaration $variableDeclaration): Variable
     {
-        return [$this->createVariableClass($variableDeclaration), 'createFromDataModel']($variableDeclaration);
+        $variableClassName = $this->createVariableClassName($variableDeclaration);
+
+        return $variableClassName::createFromDataModel($variableDeclaration);
     }
 
     /**
@@ -52,7 +54,7 @@ final class VariableFactory implements VariableFactoryInterface
      * @return string|Variable A matching Variable class
      * @throws UnexpectedValueException If $variableDeclaration is not consistent.
      */
-    private function createVariableClass(VariableDeclaration $variableDeclaration): string
+    private function createVariableClassName(VariableDeclaration $variableDeclaration): string
     {
         $variableDeclarationClass = get_class($variableDeclaration);
 
--- a/src/qtism/runtime/common/VariableFactory.php
+++ b/src/qtism/runtime/common/VariableFactory.php
@@ -54,16 +54,18 @@ final class VariableFactory implements VariableFactoryInterface
      */
     private function createVariableClass(VariableDeclaration $variableDeclaration): string
     {
-        if (!isset(self::VARIABLE_DECLARATION_MAP[$variableDeclaration::class])) {
+        $variableDeclarationClass = get_class($variableDeclaration);
+
+        if (!isset(self::VARIABLE_DECLARATION_MAP[$variableDeclarationClass])) {
             throw new UnexpectedValueException(
                 sprintf(
                     '`%s` is an unexpected `%s` implementation.',
-                    $variableDeclaration::class,
+                    $variableDeclarationClass,
                     VariableDeclaration::class
                 )
             );
         }
 
-        return self::VARIABLE_DECLARATION_MAP[$variableDeclaration::class];
+        return self::VARIABLE_DECLARATION_MAP[$variableDeclarationClass];
     }
 }
--- a/src/qtism/runtime/common/VariableFactory.php
+++ b/src/qtism/runtime/common/VariableFactory.php
@@ -1,0 +1,69 @@
+<?php
+
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ *
+ * @author Sergei Mikhailov <sergei.mikhailov@taotesting.com>
+ * @license GPLv2
+ */
+
+declare(strict_types=1);
+
+namespace qtism\runtime\common;
+
+use qtism\data\state\OutcomeDeclaration;
+use qtism\data\state\ResponseDeclaration;
+use qtism\data\state\TemplateDeclaration;
+use qtism\data\state\VariableDeclaration;
+use UnexpectedValueException;
+
+final class VariableFactory implements VariableFactoryInterface
+{
+    private const VARIABLE_DECLARATION_MAP = [
+        TemplateDeclaration::class => TemplateVariable::class,
+        OutcomeDeclaration::class  => OutcomeVariable::class,
+        ResponseDeclaration::class => ResponseVariable::class,
+    ];
+
+    /**
+     * @inheritDoc
+     */
+    public function createFromDataModel(VariableDeclaration $variableDeclaration): Variable
+    {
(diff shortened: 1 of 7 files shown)


src/qtism/runtime/common/VariableFactory.php
```diff
@@ -0,0 +1,71 @@
...
+        OutcomeDeclaration::class  => OutcomeVariable::class,
+        ResponseDeclaration::class => ResponseVariable::class,
+    ];
+
+    /**
+     * @inheritDoc
+     */
+    public function createFromDataModel(VariableDeclaration $variableDeclaration): Variable
+    {
+        return [$this->createVariableClass($variableDeclaration), 'createFromDataModel']($variableDeclaration);
```
julien-sebire, line 47: This construction seems a bit overkill to me. Why not directly do the mapping lookup in this method?
Maybe a temporary variable wouldn't hurt.

src/qtism/runtime/common/VariableFactory.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * @param VariableDeclaration $variableDeclaration A VariableDeclaration object from the QTI Data Model.
+     * @return string|Variable A matching Variable class
+     * @throws UnexpectedValueException If $variableDeclaration is not consistent.
+     */
+    private function createVariableClass(VariableDeclaration $variableDeclaration): string
```
julien-sebire, line 55: Name of this method is misleading. It returns a class name, not an instance of a class.

src/qtism/runtime/common/VariableFactoryInterface.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace qtism\runtime\common;
+
+use UnexpectedValueException;
+use qtism\data\state\VariableDeclaration;
+
+interface VariableFactoryInterface
```
julien-sebire, line 31: I understand the principle of having an interface for each class (you made the implementation final), but I really don't see another implementation to be necessary anytime soon. I'd drop the interface.

src/qtism/runtime/storage/binary/QtiBinaryStreamAccess.php
```diff
@@ -78,22 +80,30 @@ class QtiBinaryStreamAccess extends BinaryStreamAccess
...
     const RW_CORRECTRESPONSE = 2;
 
-    /**
-     * @var FileManager
-     */
+    private const VARIABLE_TYPES = [
+        'outcomeDeclaration',
+        'responseDeclaration',
+        'templateDeclaration',
+    ];
```
julien-sebire, line 87: Please add indexes. The integer values encoded in the session have fixed values (0 = outcomeVariable, 1 = responseVariable, 2 = templateVariable) and not having them in the mapping obfuscate the fact that they matter. If you don't want to rely on magic number, constants as keys are even better.

src/qtism/runtime/storage/binary/QtiBinaryStreamAccess.php
```diff
@@ -78,22 +80,30 @@ class QtiBinaryStreamAccess extends BinaryStreamAccess
...
+
     /**
      * Create a new QtiBinaryStreamAccess object.
      *
      * @param IStream $stream The IStream object to be accessed.
      * @param FileManager $fileManager The FileManager object to handle file variable.
      * @throws StreamAccessException
      */
-    public function __construct(IStream $stream, FileManager $fileManager)
+    public function __construct(IStream $stream, FileManager $fileManager, VariableFactoryInterface $variableFactory)
```
julien-sebire, line 102: This is introducing a dependency for an internal encoding issue, exposing some part of the encoding mechanism. It forces all calls to this class to create and inject a new dependency that should be internal.

This class is used outside of this repository. See:
- https://github.com/oat-sa/bundle-qti/blob/1aca1feb280c82db08538887bb0e008589865463/Accessor/TestSessionAccessor.php#L77
- https://github.com/oat-sa/extension-tao-testqti/blob/f4c615c49373432bfb259c214d73504dce4d2e39/helpers/class.TestSessionStorage.php#L290

Adding a dependency to the constructor is forcing users to change their calls. Please remember this is open source.
At worst, make it optional and instantiate the default implementation in the constructor when it's not provided.
At best, since this factory is just for internal purpose, directly instantiate it or remove the dependency (used only for a mapping lookup).

src/qtism/runtime/storage/binary/QtiBinaryStreamAccess.php
```diff
@@ -728,27 +738,25 @@ public function readAssessmentItemSession(
...
-                        $variable = $seeker->seekComponent('responseDeclaration', $varPosition);
-                        $variable = ResponseVariable::createFromDataModel($variable);
-                    } elseif ($varNature === 2) {
-                        // template
-                        $variable = $seeker->seekComponent('templateDeclaration', $varPosition);
-                        $variable = TemplateVariable::createFromDataModel($variable);
-                    }
+                    /** @var VariableDeclaration $variableDeclaration */
+                    $variableDeclaration = $seeker->seekComponent(
+                        $this->decodeVariableType($varNature),
```
julien-sebire, line 744: Variable type resolution should be done before the try loop. Then `decodeVariableType` method is not necessary, it's just a mapping lookup.

test/qtismtest/runtime/common/VariableFactoryTest.php
```diff
@@ -0,0 +1,88 @@
...
+use qtism\runtime\common\ResponseVariable;
+use qtism\runtime\common\TemplateVariable;
+use qtism\runtime\common\Variable;
+use qtism\runtime\common\VariableFactory;
+use qtismtest\QtiSmTestCase;
+
+class VariableFactoryTest extends QtiSmTestCase
+{
+    /** @var VariableFactory */
+    private $sut;
```
julien-sebire, line 44: +1 for Subject Under Test variable name! Thank you Wazelin! :)

julien-sebire (changes requested):
Thank you for taking the time to make this PR.
I get the idea of offering extensibility with the variable factory but it seems a bit overkill to me as it only deals with a mapping lookup. You can greatly simplify this and not expose any encoding to the users of this repository.

src/qtism/runtime/common/VariableFactory.php
```diff
@@ -0,0 +1,71 @@
...
+{
+    private const VARIABLE_DECLARATION_MAP = [
+        TemplateDeclaration::class => TemplateVariable::class,
+        OutcomeDeclaration::class  => OutcomeVariable::class,
+        ResponseDeclaration::class => ResponseVariable::class,
+    ];
+
+    /**
+     * @inheritDoc
+     */
```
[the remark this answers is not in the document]
wazelin: That's the _inline_ `{@inheritdoc}` you're talking about.
The one I've used (while not yet being a standard) is in RFC to explicitly mark the whole DocBlock to be copied over.

More information: https://docs.phpdoc.org/3.0/guide/guides/inheritance.html.
> Currently some applications have DocBlocks containing just the {@inheritDoc} inline tag to indicate that their complete contents should be inherited. This usage breaks with the PHPDoc Standard as summaries cannot contain inline tags and inheritance is automatic; you do not need to define a special tag for it.
>
> However, it does make clear that an element has been explicitly documented (and thus not forgotten). As such we are working to include a new (normal) tag in the PHPDoc Standard @inheritDoc that will serve that purpose.

src/qtism/runtime/common/VariableFactory.php
wazelin: I may be too big of a fan of the SRP, but to me `createFromDataModel()` is responsible for actually instantiating a `Variable`, while the other method is responsible for resolving the implementation class name.

src/qtism/runtime/storage/binary/QtiBinaryStreamAccess.php
wazelin: Yeah, I was expecting something like that.
Thanks for pointing to the exact usages!

src/qtism/runtime/common/VariableFactoryInterface.php
wazelin: I prefer having `final` default implementations exactly to enforce library users not to abuse class extension, which wouldn't be possible without an interface.

src/qtism/runtime/common/VariableFactory.php
julien-sebire: Yeah, it's all a matter of good balance between a very vertical and a very flat class tree. ;-)

m-lagarde (approved):

PR #279, opened 2021-05-27
feat: don't stop response processing on broken rule

m-lagarde:
[TR-1260](https://oat-sa.atlassian.net/browse/TR-1260)

The goal of this PR is to ignore broken process rules to skip the erroneous outcome value generation but continue the remaining ones

### DIFF
--- a/src/qtism/runtime/processing/ResponseProcessingEngine.php
+++ b/src/qtism/runtime/processing/ResponseProcessingEngine.php
@@ -33,6 +33,7 @@ use qtism\runtime\common\ProcessingException;
 use qtism\runtime\common\State;
 use qtism\runtime\rules\RuleEngine;
 use qtism\runtime\rules\RuleProcessingException;
+use qtism\runtime\rules\ProcessingCollectionException;
 
 /**
  * Class ResponseProcessingEngine
@@ -161,7 +162,7 @@ class ResponseProcessingEngine extends AbstractEngine
     public function process()
     {
         $rules = $this->getResponseProcessingRules();
-        $exception = null;
+        $processingCollectionException = null;
 
         foreach ($rules as $rule) {
             try {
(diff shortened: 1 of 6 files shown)


src/qtism/runtime/processing/ResponseProcessingEngine.php
```diff
@@ -161,19 +161,26 @@ protected function &getTemplateMapping()
...
 
-        try {
-            foreach ($rules as $rule) {
+        foreach ($rules as $rule) {
+            try {
                 $engine = new RuleEngine($rule, $this->getContext());
                 $engine->process();
                 $this->trace($rule->getQtiClassName() . ' executed');
+            } catch (\Throwable $exception) {
+                if ($exception instanceof RuleProcessingException
```
emgolubev, line 172: It will ignore other exceptions which occur earlier, right?

julien-sebire: You're right, only the last exception would be thrown. We'll lose all previous ones.

julien-sebire (changes requested):
I understand the business value of this, but I'm concerned about the outcome of this. We'll basically end up with messed up results and not having all the exceptions thrown but only the last one will prevent us to actually see what was performed and what was not...

src/qtism/runtime/processing/ResponseProcessingEngine.php
m-lagarde: Is this an issue? I mean the point of these exception is to raise the fact there was an issue during this process

Also this is already what is happening now, but instead of throwing only the last one, we throw the first one

julien-sebire: It would be better to gather all the exceptions into one specific kind of exception, so that qti-sdk is just giving all the information about what is done and what is not, and the calling code to do whatever they want with it (i.e. log the different exceptions then keep the test going or stopping). Most importantly, the calling code must be aware that the computed variables (E.G. SCORE, MAXSCORE, ...) are not relevant and scoring has to be redone afterwards with fixed rules. Having the exact feedback of every wrongly written rule is then the best help we can get from this.

julien-sebire (approved):
This is good for me. The calling code should be aware of the change in the Exception type and manage the different exceptions contained.

emgolubev (approved):
**Review Checklist**
- [ ] Acceptance criteria are respected
- [ ] Pull request title and description are meaningful
- [ ] Pull request's target is not `master`
- [ ] Commits are following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] Changelog is updated according to changes (if applicable)
- [ ] Documentation is updated according to changes (if applicable)

PR #281, opened 2021-06-03
feat: wrap processing exception into collection before to throw them

m-lagarde:
[TR-1260](https://oat-sa.atlassian.net/browse/TR-1260)

The goal of this PR is to backport the response processing behaviour on error, ignoring broken process rules to skip the erroneous outcome value generation but continue the remaining ones


julien-sebire (approved):
All good for me.

PR #288, opened 2021-07-05
Fix/tr 1605 preconditions assessmentsection level

bugalot:
### Description

This PR appears in the context of resolving AUT-753, TR-1607, TR-1605 and TR-1606. It seems this bug fix will solve them all in a row :crossed_fingers: !

Long story short, the `preConditions`, to whatever they were bound to (`assessmentSection` or `assessmentItemRef`), were always behaving being bound to an `assessentItemRef`. As a result, a `preCondition` returning true, bound to an `assessmentSection` was leading to always behave as it was bound to the first `assessmentItemRef` of the `assessmentSection`.

This is now solved at the `assessmentTestSession` instantiation level, where the `preConditions` are applied to all `routeItems` that are children to the `sectionPart` (parent class of `assessmentSection` and `assessmentItemRef`) the `preCondition` belong to.

Enjoy :sweat_smile:  !

### Appendix

**General:** 
the existing unit tests were all about `assessmentItemRef` based `preCondition`s. They continue to validate the appropriate behaviour of the routing engine.

**Hints about the changes:**
* `qtism/runtime/tests/AbstractSessionManager.php`: this is where the magic occurs. Basically, the instantiation is all about flattening the Assessment Graph (it's an N-ary tree so, yes, a Graph) while performing a Dept First Search (DFS Left -> Right) traversal. Instead of applying the `preCondition`s on the first item of the sub-routes (flatten `sectionPart`s) while going up in `sectionPart`s, we apply them on all the children. Yes, logic, but still there for 8 years :sweat_smile: !

* `test/qtismtest/runtime/tests/AssessmentTestSessionPreConditionsTest.php`: this is where the testing of the beast occurs. A first section (`S01`) contains a branchRule in order to go in a second (`S02`) OR third (`S03`) section, depending on your response. The two sections are "guarded" by `assessmentSection` level `preCondition`.

* ` test/samples/custom/runtime/preconditions/preconditions_sections_level_linear.xml`: Assessment Material to be executed to simulate the AssessmentTestSession tested in the Test Case mentioned above.

### DIFF
--- a/qtism/runtime/tests/AbstractSessionManager.php
+++ b/qtism/runtime/tests/AbstractSessionManager.php
@@ -247,8 +247,10 @@ abstract class AbstractSessionManager
                             $route->getLastRouteItem()->addBranchRules($current->getBranchRules());
 
                             // Do the same as for branch rules for pre conditions, except that they must be
-                            // attached on the first item of the route.
-                            $route->getFirstRouteItem()->addPreConditions($current->getPreConditions());
+                            // attached on all the items of the selected route.
+                            foreach ($route as $selectableRouteItem) {
+                                $selectableRouteItem->addPreConditions($current->getPreConditions());
+                            }
                         }
 
                         array_push($routeStack, $route);
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ composer.phar
 vendor
 *.geany
 *.idea
+.phpunit.result.cache
--- a/test/qtismtest/runtime/tests/AssessmentTestSessionPreConditionsTest.php
+++ b/test/qtismtest/runtime/tests/AssessmentTestSessionPreConditionsTest.php
@@ -163,4 +163,78 @@ class AssessmentTestSessionPreConditionsTest extends QtiSmAssessmentTestSessionT
 
         $this::assertFalse($testSession->isRunning());
     }
+
+    /**
+     * @throws \qtism\data\storage\php\PhpStorageException
+     * @throws \qtism\data\storage\xml\XmlStorageException
+     * @throws \qtism\runtime\tests\AssessmentItemSessionException
+     * @throws \qtism\runtime\tests\AssessmentTestSessionException
+     */
+    public function testSectionLevelByTakingS02()
+    {
+        // Expected flow: Q01, Q02, Q03
+        $testSession = self::instantiate(self::samplesDir() . 'custom/runtime/preconditions/preconditions_sections_level_linear.xml');
+        $testSession->beginTestSession();
+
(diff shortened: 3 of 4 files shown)


mike-ionut-mihai-sandu-tao (approved):
if possible let's squash the commits and use  common commits standard , 
otherwise looks good to me, 

- [ ] New code is respecting code style rules
- [ ] New code is not subject to concurrency issues (if applicable)
- [ ] Acceptance criteria are respected
- [ ] Pull request title and description are meaningful

julien-sebire (approved):
Nice catch!
Good for me.
I'd tend to consider this a non-breaking change, therefore would release as 0.24.12.

qtism/runtime/tests/AbstractSessionManager.php
```diff
@@ -247,8 +247,10 @@ protected function createRoute(AssessmentTest $test)
                             $route->getLastRouteItem()->addBranchRules($current->getBranchRules());
 
                             // Do the same as for branch rules for pre conditions, except that they must be
-                            // attached on the first item of the route.
-                            $route->getFirstRouteItem()->addPreConditions($current->getPreConditions());
+                            // attached on all the items of the selected route.
+                            foreach ($route as $selectableRouteItem) {
+                                $selectableRouteItem->addPreConditions($current->getPreConditions());
+                            }
```
uncleempty, line 253: Hey, dude, I have small proposal:
what if we add `hasPreConditions` method in both `TestPart` and `SectionPart` classes (though we only need `SectionPart` changes for the current case)? The `foreach` exporession would then be changed to smth like 
```php
if ($current->hasPreConditions()) {
    foreach ($route as $selectableRouteItem) {
        ...
    }
}
```

It would make "our" lifes a bit easier, meaning that we won't need to go through _each existing_ item in _each existing_ test sections, even if we don't have any preconditions.
How does that sound for you?

uncleempty (changes requested):
:+1: 
Good job! 
Just a small proposal from me ;)

wazelin (approved):

qtism/runtime/tests/AbstractSessionManager.php
wazelin: A good suggestion in general, but it won't really make any difference in terms of complexity or performance. In fact it will increase the cyclomatic complexity in the end.

uncleempty: I know, I was just testing it in local, and was a bit tired to go through this line on big sections with no preconditions and a lot of items while debugging. that's why decided to propose :)

bugalot: I will follow @wazelin on this one. Tend to agree. Thanks @uncleempty for the proposal!

PR #290, opened 2021-07-06
Fixed security issue on league/flysystem.

julien-sebire:
This PR fixes a security issue found in league/flysystem dependency.
See https://github.com/oat-sa/qti-sdk/security/dependabot/composer.json/league%2Fflysystem/open for more information.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
 		"ext-libxml": "*",
 		"ext-mbstring": "*",
 		"ext-simplexml": "*",
-		"league/flysystem": "^1.0",
+		"league/flysystem": ">=1.1.4 <2.0.0",
 		"oat-sa/lib-beeme": "0.2.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -31,7 +31,7 @@
 		"ext-libxml": "*",
 		"ext-mbstring": "*",
 		"ext-simplexml": "*",
-		"league/flysystem": "^1.0",
+		"league/flysystem": ">=1.1.4 <2.0.0",
```
wazelin, line 34: A more clean and concise equivalent would be the following.

```suggestion
		"league/flysystem": "^1.1.4",
```

wazelin (changes requested):

augustas (approved):

julien-sebire:
This is not needed any more because the fix is applied with the current constraint.

PR #291, opened 2021-07-12
feat: value marshaller enclose content in CDATA section

augustas:
XML handling would break if xml not compatible html characters are used, like `&nbsp;`.
 
Related to : https://oat-sa.atlassian.net/browse/TR-1412

XML storage Value marshaller updated to enclose content in CDATA section.
  
#### How to test
 - author a test with Extended text interaction
 - publish the test and launch the delivery as a test taker
 - answer extended text interaction item with content containing non breaking spaces ` `, finish the test
 - verify that result extraction consumer processed results without failure

### DIFF
--- a/src/qtism/data/storage/xml/marshalling/ValueMarshaller.php
+++ b/src/qtism/data/storage/xml/marshalling/ValueMarshaller.php
@@ -23,12 +23,14 @@
 
 namespace qtism\data\storage\xml\marshalling;
 
+use DOMCdataSection;
 use DOMElement;
 use InvalidArgumentException;
 use qtism\common\enums\BaseType;
 use qtism\data\QtiComponent;
 use qtism\data\state\Value;
 use qtism\data\storage\Utils;
+use qtism\data\storage\xml\Utils as XmlUtils;
 
 /**
  * Marshalling/Unmarshalling implementation for value.
@@ -97,7 +99,7 @@ class ValueMarshaller extends Marshaller
         $fieldIdentifer = $component->getFieldIdentifier();
         $baseType = $component->getBaseType();
 
-        self::setDOMElementValue($element, $component->getValue());
+        $element->appendChild($this->encloseValueInCDATASection($component->getValue()));
 
         if (!empty($fieldIdentifer)) {
             $this->setDOMElementAttribute($element, 'fieldIdentifier', $fieldIdentifer);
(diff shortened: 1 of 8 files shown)


src/qtism/data/storage/xml/marshalling/ValueMarshaller.php
```diff
@@ -164,4 +167,30 @@ public function getExpectedQtiClassName()
...
+        $valueChildNodes = self::getChildElements($element, true);
+        if (!empty($valueChildNodes)
+            && ($node = reset($valueChildNodes))
+            && $node->nodeType === XML_CDATA_SECTION_NODE
+        ) {
+            return trim($node->nodeValue);
+        } else {
+            return trim($element->nodeValue);
+        }
+    }
```
wazelin, line 195: `reset($valueChildNodes)` cannot be empty or undefined if `!empty($valueChildNodes)`. Besides `$node->nodeType === XML_CDATA_SECTION_NODE` should always be `true`. Otherwise, something went very wrong.
Also, let's separate the concerns just a tiny bit. WDYT of the following?

```suggestion
        $valueNode = $this->extractValueNode($element);

        return trim($valueNode->nodeValue);
    }

    private function extractValueNode(DOMElement $element): DOMElement
    {
        $valueChildNodes = self::getChildElements($element, true);

        foreach(self::getChildElements($element, true) as $valueNode) {
            if ($valueNode->nodeType === XML_CDATA_SECTION_NODE) {
                return $valueNode;
            }
        }

        return $element;
    }
```

emgolubev:
@augustas please check @wazelin suggestions, I think it makes sense, other code looks and works fine. Good job.

src/qtism/data/storage/xml/marshalling/ValueMarshaller.php
augustas: > reset($valueChildNodes) cannot be empty or undefined if !empty($valueChildNodes)

Yes, that is what I expect, I use it to pick the first found node and create a new variable at the same time. 

> Besides $node->nodeType === XML_CDATA_SECTION_NODE should always be true.

No, not necessarily. Since loaded XML values might not have the CDATA wrapping, to make it a non-breaking change I suggest for `<value>` element to accept text or CDATA wrapped text.

wazelin: > No, not necessarily.

But it is, when taken from `self::getChildElements()`.

> to make it a non-breaking change I suggest for <value> element to accept text or CDATA wrapped text

Yeah, that's clear :)
I just suggest to separate the value-containing node extraction logic from the value extraction itself.

augustas: I tested that returned child nodes would return both `DOMText` and `DOMCdataSection` nodes, so the code could be simplified

wazelin (approved):

bugalot:
Hello guys!

I am sorry to disagree with that fix. The core issue is that SimpleXml breaks with nbsp. We are know impacting the Assessment Materials (items) we produce to circumvent a SimpleXml weakness. It cannot be like that, we have to keep the implementation clean.

I strongly suggest we remove the usage of SimpleXml in `qtism\data\storage\xml\Utils::findExternalNamespaces` to properly fix the issue. Sorry to put my veto here.

bugalot (changes requested):
See comments

wazelin:
I see your point, @bugalot, but I can't really agree with it.

A non-escaped `&nbsp;` is an invalid XML entity, and this is what we should be fixing. SimpleXml fails to parse the XML for quite a valid reason.

wazelin (changes requested):
An alternative solution is available in #293.

Consider closing this PR.

bugalot:
OK I got the whole picture now @wazelin . I cannot say anymore that this PR is "invalid". I'm just not sure it's the best way to solve TR-1412. I find https://github.com/oat-sa/qti-sdk/pull/293 way more elegant.

However, I can see that the use of CDATA is in the acceptance criteria... You will have guys to debate this without me :smile:

PR #296, opened 2021-08-03
feat: allow item jump on linear test when manually set

m-lagarde:
[TR-1264](https://oat-sa.atlassian.net/browse/TR-1264)

Allow jump navigation on test session for linear tests when manually set

### DIFF
--- a/src/qtism/runtime/tests/AssessmentTestSession.php
+++ b/src/qtism/runtime/tests/AssessmentTestSession.php
@@ -197,6 +197,15 @@ class AssessmentTestSession extends State
     private $config = 0;
 
     /**
+     * Whether or not allowing jump in any case.
+     *
+     * If enabled, jumps will be allowed even if the current navigation mode is linear.
+     *
+     * @var bool
+     */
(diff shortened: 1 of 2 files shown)


julien-sebire (approved):
Ok, provided that this doesn't lead to the navigation mode to be non-linear upon re-launched test.

src/qtism/runtime/tests/AssessmentTestSession.php
```diff
@@ -677,7 +698,7 @@ public function mustTrackPath()
      */
     public function mustAlwaysAllowJumps()
     {
-        return (bool)($this->getConfig() & self::ALWAYS_ALLOW_JUMPS);
+        return (bool)($this->getConfig() & self::ALWAYS_ALLOW_JUMPS) || $this->alwaysAllowJumps;
```
wazelin, line 701: The first condition is a bit heavier. I recommend putting it to the end of the logical `or`.

wazelin (changes requested):

PR #304, opened 2022-01-12
fix: simplifies exceptions management

boajer:
https://oat-sa.atlassian.net/browse/TR-1265

### DIFF
--- a/src/qtism/common/ContentPackageExceptionInterface.php
+++ b/src/qtism/common/ContentPackageExceptionInterface.php
@@ -8,27 +8,24 @@
  *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
  * along with this program; if not, write to the Free Software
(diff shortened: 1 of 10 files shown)


src/qtism/common/ContentPackageExceptionInterface.php
```diff
@@ -0,0 +1,31 @@
...
+ * @license GPLv2
+ */
+declare(strict_types=1);
+
+namespace qtism\common;
+
+/**
+ * The ContentPackageExceptionInterface intended to help unify the exception handlers behavior
+ */
+interface ContentPackageExceptionInterface
```
wazelin, line 29:
```suggestion
interface PackageContentExceptionInterface
```

wazelin (changes requested):

src/qtism/common/ContentPackageExceptionInterface.php
boajer: @wazelin Should https://www.php-fig.org/bylaws/psr-naming-conventions/ be respected?

quintanilhar: What do you think about using the lib name in the base exception class just like [Guzzle does](https://github.com/guzzle/guzzle/blob/master/src/Exception/GuzzleException.php)?

Something like: `QtiSdkException`

quintanilhar (commented):
I think a good standard is to use the lib name for the base exception to make it easy to track it down.

quintanilhar (approved):
LGTM 👍🏻

julien-sebire (approved):
All good for me. Nice job!

PR #315, opened 2022-06-13
fix: legacy : TR-4282 remove extra method; data loaded by rules no need in tha…

kilatib:
[TR-4282](https://oat-sa.atlassian.net/browse/TR-4282)

Looks like that if is overhead for that case, Little down in a code we see that we load rules and no rules. If element required external score we don't have rules for it, so logic working correct.

--------

[DEMO VIDEO ](https://www.loom.com/share/3c720f3dc6184d069f97e82928ceae42)


qtism/runtime/tests/AssessmentItemSession.php
```diff
@@ -748,11 +712,11 @@ public function endAttempt(State $responses = null, $responseProcessing = true,
                 $this->resetOutcomeVariables();
             }
 
-            $responseProcessing = $this->assessmentItem->getResponseProcessing();
```
kilatib, line 751: I think it was bad idea to use variable with same name as in method params

PR #317, opened 2022-06-28
Fix/tr 4332/patternmask length validation

oatymart:
## [TR-4332](https://oat-sa.atlassian.net/browse/TR-4332): Incorrect maxLength validation for non-latin characters in Inline text

This bug occurs when an TextEntry interaction is authored with a maximum length, resulting in the `patternMask` attribute being a regex `^[\\s\\S]{0,N}$` used for input length validation. Today, that validation is applied differently on backend and Solar frontend, which [parses](https://github.com/oat-sa/tao-deliver-fe/blob/develop/packages/item-runner/src/runner/util/patternMask.js#L31) the `maxlength` number back out of the pattern in order to check the input's length does not exceed it.

This is significant on Unicode inputs (see [byte counter](https://mothereff.in/byte-counter)):
- `a` // JS string length = 1; bytes = 1; patternMask `{0,1}` = valid
- `å` // JS string length = 1; bytes = 2; patternMask `{0,1}` = invalid
- `😊` // JS string length = 1; bytes = 4; patternMask `{0,1}` = invalid

This solution (kind of) matches backend's behaviour to frontend's by adding the Unicode `u` modifier to the BE regex, meaning that all of the above inputs match as 1 character. Therefore no input should be able to pass FE validation but fail on BE. (It could also be done by parsing out the integer and discarding the `patternMask`, if desired.)

### To test:
- run the unit test
- incorporate this into `tao-deliver-be` and take the test from the Jira ticket
  - see that Unicode inputs are validated the same on BE as FE

Screenshot 2022-06-28 at 09 50 31

### DIFF
--- a/src/qtism/runtime/expressions/operators/Utils.php
+++ b/src/qtism/runtime/expressions/operators/Utils.php
@@ -350,10 +350,16 @@ class Utils
         return $pattern;
     }
 
-     private static function withoutStringAnchors(string $pattern): string
-     {
-         $pattern = ltrim($pattern, '^');
+    /**
+     * Remove start and end anchors from a pattern.
+     *
+     * @param string $pattern
+     * @return string
+     */
+    private static function withoutStringAnchors(string $pattern): string
+    {
+        $pattern = ltrim($pattern, '^');
 
-         return rtrim($pattern, '$');
-     }
+        return rtrim($pattern, '$');
+    }
 }
--- a/src/qtism/runtime/expressions/operators/Utils.php
+++ b/src/qtism/runtime/expressions/operators/Utils.php
@@ -331,23 +331,25 @@ class Utils
      * Prepare an XSD Regular Expression pattern into a PCRE compliant one.
      *
      * @param string $pattern
+     * @param string $pcreFlags
      * @return string
      */
-     public static function prepareXsdPatternForPcre(string $pattern): string
-     {
-         // XML schema always implicitly anchors the entire regular expression
-         // Neither caret (^) nor dollar ($) sign have special meaning so they are
-         // considered as normal characters.
-         // see http://www.regular-expressions.info/xml.html
-         $pattern = self::withoutStringAnchors($pattern);
-         $pattern = self::escapeSymbols($pattern, ['$', '^']);
-         $pattern = self::pregAddDelimiter('^' . $pattern . '$');
+    public static function prepareXsdPatternForPcre(string $pattern, string $pcreFlags = ''): string
+    {
+        // XML schema always implicitly anchors the entire regular expression
+        // Neither caret (^) nor dollar ($) sign have special meaning so they are
+        // considered as normal characters.
+        // see http://www.regular-expressions.info/xml.html
+        $pattern = self::withoutStringAnchors($pattern);
+        $pattern = self::escapeSymbols($pattern, ['$', '^']);
+        $pattern = self::pregAddDelimiter('^' . $pattern . '$');
 
-         // XSD regexp always case-sensitive (nothing to do), dot matches white-spaces (use PCRE_DOTALL).
-         $pattern .= 's';
+        // XSD regexp always case-sensitive (nothing to do), dot matches white-spaces (use PCRE_DOTALL).
+        $pattern .= 's';
+        $pattern .= $pcreFlags;
 
-         return $pattern;
-     }
+        return $pattern;
+    }
 
      private static function withoutStringAnchors(string $pattern): string
      {
--- a/src/qtism/runtime/expressions/operators/Utils.php
+++ b/src/qtism/runtime/expressions/operators/Utils.php
@@ -334,7 +334,7 @@ class Utils
      * @param string $pcreFlags
      * @return string
      */
-    public static function prepareXsdPatternForPcre(string $pattern, string $pcreFlags = ''): string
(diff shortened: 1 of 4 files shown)


src/qtism/runtime/tests/Utils.php
```diff
@@ -84,7 +84,13 @@ public static function isResponseValid(QtiDatatype $response = null, ResponseVal
                 $values = ($cardinality === Cardinality::SINGLE) ? [$response->getValue()] : $response->getArrayCopy();
             }
 
-            $patternMask = OperatorUtils::prepareXsdPatternForPcre($patternMask);
+            // Is the patternMask a length validator or not?
+            // If it is, apply unicode flag - this will prevent a mismatch between frontend and backend on the string lengths counted as valid
+            $lengthPattern = '/\^\[\\\\s\\\\S\]\{0,\d+\}\$/';
+            $isLengthValidator = @preg_match($lengthPattern, $patternMask) === 1;
+            $pcreFlags = $isLengthValidator ? 'u' : '';
```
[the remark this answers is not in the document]
oatymart: I guess you are right, I was trying to limit the scope of the changes for regressions, but it does sound perfectly harmless.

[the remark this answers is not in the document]
wazelin: In fact, what if we just added the flag by default in `prepareXsdPatternForPcre()`? :)
I mean, the QTI (XML) value is a string of unicode characters by its nature. So, it's only expected to compare it against a pattern with a unicode flag applied.

src/qtism/runtime/expressions/operators/Utils.php
```diff
@@ -331,23 +331,25 @@ public static function lastPregErrorMessage()
...
+        // Neither caret (^) nor dollar ($) sign have special meaning so they are
+        // considered as normal characters.
+        // see http://www.regular-expressions.info/xml.html
+        $pattern = self::withoutStringAnchors($pattern);
+        $pattern = self::escapeSymbols($pattern, ['$', '^']);
+        $pattern = self::pregAddDelimiter('^' . $pattern . '$');
+
+        // XSD regexp always case-sensitive (nothing to do), dot matches white-spaces (use PCRE_DOTALL).
+        $pattern .= 's';
+        $pattern .= $pcreFlags;
```
wazelin, line 349: I believe the `u` flag is better to be applied by default in all the cases. The values the pattern is applied to are unicode after all.

```suggestion
    public static function prepareXsdPatternForPcre(string $pattern, string $pcreFlags = 'su'): string
    {
        // XML schema always implicitly anchors the entire regular expression
        // Neither caret (^) nor dollar ($) sign have special meaning so they are
        // considered as normal characters.
        // see http://www.regular-expressions.info/xml.html
        $pattern = self::withoutStringAnchors($pattern);
        $pattern = self::escapeSymbols($pattern, ['$', '^']);
        $pattern = self::pregAddDelimiter('^' . $pattern . '$');

        // XSD regexp always case-sensitive (nothing to do), dot matches white-spaces (use PCRE_DOTALL).
        $pattern .= $pcreFlags;
```

wazelin (changes requested):

oatymart:
Hold on while I repair other unit tests I forgot to check.

src/qtism/runtime/expressions/operators/Utils.php
```diff
@@ -331,23 +331,24 @@ public static function lastPregErrorMessage()
...
+        // see http://www.regular-expressions.info/xml.html
+        $pattern = self::withoutStringAnchors($pattern);
+        $pattern = self::escapeSymbols($pattern, ['$', '^']);
+        $pattern = self::pregAddDelimiter('^' . $pattern . '$');
+
+        // XSD regexp always case-sensitive (nothing to do), dot matches white-spaces (use PCRE_DOTALL).
+        $pattern .= $pcreFlags;
+
+        return $pattern;
+    }
```
wazelin, line 351: Hm, something went sideways with the indentation.
```suggestion
     public static function prepareXsdPatternForPcre(string $pattern, string $pcreFlags = 'su'): string
     {
         // XML schema always implicitly anchors the entire regular expression
         // Neither caret (^) nor dollar ($) sign have special meaning so they are
         // considered as normal characters.
         // see http://www.regular-expressions.info/xml.html
         $pattern = self::withoutStringAnchors($pattern);
         $pattern = self::escapeSymbols($pattern, ['$', '^']);
         $pattern = self::pregAddDelimiter('^' . $pattern . '$');

         // XSD regexp always case-sensitive (nothing to do), dot matches white-spaces (use PCRE_DOTALL).
         $pattern .= $pcreFlags;

         return $pattern;
     }
```

wazelin (changes requested):
Looks good!

Just a slight indentation issue on one of the files.

src/qtism/runtime/expressions/operators/Utils.php
oatymart: Are you sure? When I first went in, this method had weird `4n+1` indentation, I only reformatted it to 4.

wazelin: Sorry for that, totally my bad here. All good with the PR!

wazelin (approved):

kilatib (approved):

PR #319, opened 2022-07-12
feat: Add serialization support to SerializableDomDocument

kilatib:
# [TR-4391](https://oat-sa.atlassian.net/browse/TR-4391)

1. Add Serialization to SerializableDomDocument
2. Increase used version of PHP to support magic methods: __serialize(), __unserialize(array $data)


kilatib:
not all test ready need to be procced more

PR #320, opened 2022-07-13
Feature TR-4391 ensure PHP 8.1 support

kilatib:
# [TR-4391](https://oat-sa.atlassian.net/browse/TR-4391)

## Development impact 
1. BREAKING CHANGE: `SerializableDomDocument` no longer extends `DOMDocument` due to a serialization-preventing bug in PHP 8.1 https://github.com/php/php-src/issues/8996
2. BREAKING CHANGE:  Drop support for PHP <7.4
3. Now where it possible  after call getXml() we should assign `ExternalQtiComponent`  as `SerializableDomDocument`  not like `DOMDocument`
4. Correct test to check right class type.

### DIFF
--- a/src/qtism/common/dom/SerializableDomDocument.php
+++ b/src/qtism/common/dom/SerializableDomDocument.php
@@ -36,7 +36,7 @@ use DOMAttr;
 use DOMEntityReference;
 use DOMNode;
 use DOMNodeList;
-use Exception;
+use Error;
 
 /**
  * Serializable DOM Document
@@ -111,9 +111,9 @@ class SerializableDomDocument
     public function __serialize(): array
     {
         return [
-            'version'  => $this->dom->xmlVersion,
-            'encoding' => $this->dom->encoding,
-            'xmlData'  => (string)$this->__toString()
+            'version'  => (string)$this->dom->xmlVersion,
+            'encoding' => (string)$this->dom->encoding,
+            'xmlData'  => (string)$this,
         ];
     }
 
@@ -125,16 +125,14 @@ class SerializableDomDocument
 
     public function __toString(): string
     {
-        return $this->dom->saveXML();
+        $xml = $this->dom->saveXML();
+        return $xml ? : '';
     }
 
-    /**
-     * @throws Exception
-     */
     public function __call($name, $arguments)
     {
         if (!method_exists($this->dom, $name)) {
-            throw new Exception(__CLASS__ . 'has no method::' . $name);
+            throw new Error(sprintf('Call to undefined method %s::%s()', __CLASS__, $name));
         }
 
         return call_user_func_array([$this->dom, $name], $arguments);
@@ -143,7 +141,7 @@ class SerializableDomDocument
     public function __get($name)
     {
         if (!property_exists($this->dom, $name)) {
-            throw new Exception(__CLASS__ . 'has no property::' . $name);
+            trigger_error(sprintf('Undefined property: %s::%s', __CLASS__, $name), E_USER_WARNING);
         }
 
         return $this->dom->$name ?? null;
@@ -151,9 +149,6 @@ class SerializableDomDocument
 
     public function __set($name, $value)
     {
-        if (!property_exists($this->dom, $name)) {
-            throw new Exception(__CLASS__ . 'has no property::' . $name);
-        }
         $this->dom->$name = $value;
 
         return $this->dom;
--- a/src/qtism/common/dom/SerializableDomDocument.php
+++ b/src/qtism/common/dom/SerializableDomDocument.php
@@ -129,7 +129,7 @@ class SerializableDomDocument
         return $xml ? : '';
     }
 
-    public function __call($name, $arguments)
+    public function __call(string $name, array $arguments)
     {
         if (!method_exists($this->dom, $name)) {
             throw new Error(sprintf('Call to undefined method %s::%s()', __CLASS__, $name));
@@ -138,7 +138,7 @@ class SerializableDomDocument
         return call_user_func_array([$this->dom, $name], $arguments);
     }
 
-    public function __get($name)
+    public function __get(string $name)
     {
         if (!property_exists($this->dom, $name)) {
             trigger_error(sprintf('Undefined property: %s::%s', __CLASS__, $name), E_USER_WARNING);
@@ -147,10 +147,17 @@ class SerializableDomDocument
         return $this->dom->$name ?? null;
     }
 
-    public function __set($name, $value)
+    public function __set(string $name, $value): void
     {
         $this->dom->$name = $value;
+    }
 
-        return $this->dom;
+    public function __isset(string $name): bool
+    {
+        return isset($this->dom->$name);
+    }
+    public function __unset(string $name): void
+    {
+        unset($this->dom->$name);
     }
 }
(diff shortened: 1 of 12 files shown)


src/qtism/common/dom/SerializableDomDocument.php
```diff
@@ -24,27 +24,71 @@
 namespace qtism\common\dom;
 
 use DOMDocument;
+use Exception;
 
 /**
  * Serializable DOM Document
  *
  * This class is a PHP Serializable DOMDocument implementation.
+ * @serializable
```
wazelin, line 33: I'm not familiar with such an annotation. Is it part of some standard?

src/qtism/common/dom/SerializableDomDocument.php
```diff
@@ -24,27 +24,71 @@
...
+    {
+        return $this->dom->saveXML();
+    }
 
     /**
-     * @return array
+     * @throws Exception
      */
-    public function __sleep()
+    public function __call($name, $arguments)
```
wazelin, line 67: Adding some `@method` and `@property` annotations describing the ones available in DOMDocument would help. Let's have some on the class level.

kilatib: Done added to class head

src/qtism/common/dom/SerializableDomDocument.php
```diff
@@ -24,27 +24,138 @@
...
+        return $this->dom->$name ?? null;
     }
 
-    public function __wakeup()
+    public function __set($name, $value)
     {
-        $this->loadXML($this->xmlData);
+        if (!property_exists($this->dom, $name)) {
+            throw new Exception(__CLASS__ . 'has no property::' . $name);
+        }
```
wazelin, line 156: PHP, sadly, allows dynamic properties.

This will be more aligned with a real-world behavior.

```suggestion
```

wazelin (changes requested):

src/qtism/common/dom/SerializableDomDocument.php
```diff
@@ -24,27 +24,138 @@
...
+
+    public function __unserialize(array $data): void
+    {
+        $this->dom = new DOMDocument($data['version'], $data['encoding']);
+        $this->dom->loadXML($data['xmlData']);
+    }
+
+    public function __toString(): string
+    {
+        return $this->dom->saveXML();
```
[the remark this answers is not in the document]
gabrielfs7: Should we mask the error in this case, then?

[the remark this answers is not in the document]
wazelin: We will not, it will still generate any notices it might want.
The point is – the underlying method return type is not strictly `string`.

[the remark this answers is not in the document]
gabrielfs7: What I mean is that, if it return `false` in case of error, should not we throw an exception with the error instead of returning empty string?

src/qtism/common/dom/SerializableDomDocument.php
```diff
@@ -24,27 +24,138 @@
...
+    {
+        $this->dom = new DOMDocument($version, $encoding);
+    }
+
+    public function __serialize(): array
+    {
+        return [
+            'version'  => $this->dom->xmlVersion,
+            'encoding' => $this->dom->encoding,
+            'xmlData'  => (string)$this->__toString()
```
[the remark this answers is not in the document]
gabrielfs7: I personally prefer `$this->__toString()` to leverage from IDE to search where the __toString method is used. Just personal opinion.

src/qtism/common/dom/SerializableDomDocument.php
```diff
@@ -24,27 +24,138 @@
...
+
+    public function __construct(string $version = '1.0', string $encoding = '')
+    {
+        $this->dom = new DOMDocument($version, $encoding);
+    }
+
+    public function __serialize(): array
+    {
+        return [
+            'version'  => $this->dom->xmlVersion,
```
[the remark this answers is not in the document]
wazelin: This doesn't seem to be resolved, @kilatib.

src/qtism/common/dom/SerializableDomDocument.php
```diff
@@ -24,27 +24,133 @@
...
+        }
+
+        return call_user_func_array([$this->dom, $name], $arguments);
+    }
 
-    /**
-     * @return array
-     */
-    public function __sleep()
+    public function __get($name)
```
wazelin, line 141: Just a suggestion.

```suggestion
    public function __get(string $name)
```

src/qtism/common/dom/SerializableDomDocument.php
```diff
@@ -24,27 +24,133 @@
...
+        $this->dom->loadXML($data['xmlData']);
+    }
+
+    public function __toString(): string
+    {
+        $xml = $this->dom->saveXML();
+        return $xml ? : '';
+    }
+
+    public function __call($name, $arguments)
```
wazelin, line 132: Just a suggestion

```suggestion
    public function __call(string $name, array $arguments)
```

src/qtism/common/dom/SerializableDomDocument.php
```diff
@@ -24,27 +24,133 @@
...
-        return ['xmlData'];
+        if (!property_exists($this->dom, $name)) {
+            trigger_error(sprintf('Undefined property: %s::%s', __CLASS__, $name), E_USER_WARNING);
+        }
+
+        return $this->dom->$name ?? null;
     }
 
-    public function __wakeup()
+    public function __set($name, $value)
```
wazelin, line 150: Just a suggestion

```suggestion
    public function __set(string $name, $value): void
```

src/qtism/common/dom/SerializableDomDocument.php
```diff
@@ -24,27 +24,133 @@
...
     {
-        $this->loadXML($this->xmlData);
+        $this->dom->$name = $value;
+
+        return $this->dom;
     }
```
wazelin, line 155: For the sake of feature-completeness in terms of property-access API, I strongly advise adding `__isset()` and `__unset()` methods as well.
```suggestion
    }

    public function __isset(string $name): bool
    {
        return isset($this->dom->$name);
    }

    public function __unset(string $name): void
    {
        unset($this->dom->$name);
    }
```

wazelin (changes requested):

src/qtism/common/dom/SerializableDomDocument.php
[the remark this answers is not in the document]
kilatib: @wazelin, `version` is deprecated we should use `xmlVersion` instead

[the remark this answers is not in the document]
wazelin: Gotcha.

src/qtism/common/dom/SerializableDomDocument.php
kilatib: done

wazelin (changes requested):
Excellent job!
Now, let's have the new magic methods covered by simple unit tests.
1. Accessing a property
2. Accessing an inexistent property, and expecting an error
3. Setting a property
4. Checking if property exists
5. Unsetting a property
6. Calling a method

kilatib:
@wazelin tests added except 
> 5. Unsetting a property

It not working proper for `DOMDocument` object its property mostly is readonly

wazelin (changes requested):
> It not working proper for `DOMDocument` object its property mostly is readonly

All right. Let's remove __unset then. It makes no sense in having it if all the state is read-only indeed. I didn't realize that.

wazelin (approved):

PR #323, opened 2022-08-12
fix: Store Item Limits to DurationStore

kilatib:
# Issue found [here](https://oat-sa.atlassian.net/browse/TR-4418)

##  Description
During creating DurationStore for Assessment Session we skip to add duration for AsessmentItem it happens because `$access`  follow to elements inside items not on its properties. 

## Development impact
1. Extend duration store with item limits data if it set for item. 

## Demo 
https://www.loom.com/share/f6d938a834754daeae6f83d884db6fb1

### DIFF
--- a/src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
+++ b/src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
@@ -341,7 +341,7 @@ abstract class AbstractQtiBinaryStorage extends AbstractStorage
                 }
 
             $item = $assessmentTestSession->getCurrentAssessmentItemRef();
-            if ($timeLimits = $item->getTimeLimits()) {
+            if ($item && $timeLimits = $item->getTimeLimits()) {
                 if ($timeLimits->hasMaxTime()) {
                     $durationVariable = new OutcomeVariable(
                         $item->getIdentifier(),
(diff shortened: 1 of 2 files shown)


poyuki (approved):
Didn't  test it but your video verification looks very convincing

src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
```diff
@@ -340,7 +340,31 @@ public function retrieve($sessionId)
                     $durationStore->setVariable($durationVariable);
                 }
 
-                $assessmentTestSession->setDurationStore($durationStore);
```
wazelin, line 343: Minor stuff, but let's fix the indentation of the lines above as well.

src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
```diff
@@ -340,7 +340,31 @@ public function retrieve($sessionId)
...
+            $item = $assessmentTestSession->getCurrentAssessmentItemRef();
```
wazelin, line 343: I'm not sure about that approach.
The purpose of this method is to read and unmarshal data as-is. If some data is lost during marshaling, it should then probably be fixed there.

wazelin (changes requested):

src/qtism/runtime/storage/binary/AbstractQtiBinaryStorage.php
kilatib: The problem is that it does marshaling by inner data of item, in that case, it skips to read time limits and it not included it to duration storage, so this block is cover that missing

wazelin: That's all right. I mean, you have already replaced the data-storage usage with currentAssessmentItemRef's max time in timer-init time.

Let's revert these changes.

wazelin (changes requested):

PR #324, opened 2022-08-22
Fix/TR-4212/Decode HTML entities in attributes

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TR-4212

To prevent XML validation errors, the XML attributes are escaped so that some entities are encoded (`"`, `&`, `<`, `>`). This is made implicitly by the PHP built-in engine, but it is also enforced by the [attribute setter](https://github.com/oat-sa/qti-sdk/blob/5119fc6c91e3e39c36a035eaef9ee840fb573320/src/qtism/data/storage/xml/Utils.php#L353). This results in double encoding. Unfortunately, it does not look like the unescape is applied properly while reading back the XML.

~~This code change adds decoding of the HTML special chars when reading attributes.~~
This code change removes the double encoding of attributes.

How to test:
- run the unit tests (`vendor/bin/phpunit`)
- check out the branch in a TAO instance (composer: `"qtism/qtism": "dev-fix/TR-4212/decode-html-entities-in-attributes as 16.6-dev",`)
- create/update a test, making sure to introduce social chars in the test title as well as in a section title. For example, add the sequence `<&>` to the tile.
- publish the test
- launch the test and look at the titles, they should not show encoded entities.

### DIFF
--- a/src/qtism/data/storage/xml/Utils.php
+++ b/src/qtism/data/storage/xml/Utils.php
@@ -284,7 +284,7 @@ class Utils
 
         switch ($datatype) {
             case 'string':
-                return $attr;
+                return htmlspecialchars_decode($attr);
 
             case 'integer':
                 return (int)$attr;
--- a/src/qtism/data/storage/xml/Utils.php
+++ b/src/qtism/data/storage/xml/Utils.php
@@ -284,7 +284,7 @@ class Utils
 
         switch ($datatype) {
             case 'string':
-                return htmlspecialchars_decode($attr);
+                return $attr;
 
             case 'integer':
                 return (int)$attr;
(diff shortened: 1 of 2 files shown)


src/qtism/data/storage/xml/Utils.php
```diff
@@ -284,7 +284,7 @@ public static function getDOMElementAttributeAs(DOMElement $element, string $att
 
         switch ($datatype) {
             case 'string':
-                return $attr;
+                return htmlspecialchars_decode($attr);
```
gabrielfs7, line 287: Since this code will decode all the attributes now, I am curious if there is any case where we, on purpose, would like to keep the attribute value encoded.

Maybe would be nice to check with @julien-sebire too.

gabrielfs7 (approved):
- [ ] New code is covered by tests (if applicable)
- [ ] New code is respecting code style rules
- [ ] New code is not subject to concurrency issues (if applicable)

hectoras (approved):

src/qtism/data/storage/xml/Utils.php
jsconan: Fixed by the new approach where we are not double encoding the attributes as it is implicitly performed by PHP through DOMElement.

julien-sebire (approved):
Brilliant! All good for me.

PR #328, opened 2022-09-06
Max word count restriction validation

poyuki:
# [TR-4539](https://oat-sa.atlassian.net/browse/TR-4539)

## Summary
pattern mask provide for response validation in case of long sequences check can fail on regex compilation step. Added another approach of such cases validation with parsing regex and extracting necessary restriction for PHP level check

## How to test
* install in deliver-be
* use test with max word constraints

## Required for 
https://github.com/oat-sa/tao-deliver-be/pull/971


src/qtism/runtime/tests/Utils.php
```diff
@@ -86,13 +89,24 @@ public static function isResponseValid(QtiDatatype $response = null, ResponseVal
...
             foreach ($values as $value) {
-                $result = @preg_match($patternMask, $value);
+                if ($isMaxEntryRestriction) {
+                    extract($matches);
+                    $entries = count(preg_split("/$splitPattern/", $response)) - 1;
+                    if ($entries > $max || $entries < $min) {
+                        return false;
+                    }
+                } else {
+                    $result = @preg_match($patternMask, $value);
```
wazelin, line 103: WDYT we try regular `preg_match()` first, and then if failed with an error, try the new approach?
I think the regular preg_match() covers the majority of the cases, and therefore should be default.

src/qtism/runtime/tests/Utils.php
```diff
@@ -125,4 +139,10 @@ public static function isResponseValid(QtiDatatype $response = null, ResponseVal
 
         return true;
     }
+
+    private static function isSingleMatchGroup(string $patternMask): bool
+    {
+        $closeBracketPosition = strpos($patternMask, self::CLOSE_MATCH_GROUP_TOKEN);
```
wazelin, line 145: If it's not too much of a trouble, could you please add a few test cases that would check this condition?

wazelin (changes requested):

wazelin (changes requested):

src/qtism/runtime/tests/Utils.php
```diff
@@ -36,8 +36,12 @@
  */
 class Utils
 {
+    private const MAX_ENTRY_RESTRICTION_PATTERN = '/^\/\^(?P<splitPattern>\([^{]+)\{(?P<min>\d+),(?P<max>\d+)\}\$\/\w*$/';
```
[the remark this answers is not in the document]
poyuki: looks great, will try to apply it. Just small doubt is it worth it make it so intellectual :)

test/qtismtest/runtime/tests/TestUtilsTest.php
```diff
@@ -105,6 +105,36 @@ public function isResponseValidProvider()
...
+                new ResponseValidityConstraint('RESPONSE', 1, 1, '^(?:(?:[^\s\:\!\?\;\…\€]+)[\s\:\!\?\;\…\€]*){1000}$')
+            ], //100 words
```
wazelin, line 137: Just a nitpick :)
```suggestion
            ], //1000 words
```

wazelin (changes requested):

src/qtism/runtime/tests/Utils.php
```diff
@@ -86,13 +89,24 @@ public static function isResponseValid(QtiDatatype $response = null, ResponseVal
...
             foreach ($values as $value) {
                 $result = @preg_match($patternMask, $value);
 
                 if ($result === 0) {
                     return false;
                 } elseif ($result === false) {
-                    throw new RuntimeException(OperatorUtils::lastPregErrorMessage());
+                    if ($isMaxEntryRestriction) {
+                        [$splitPattern, $min, $max] = self::extractMaxEntryRestrictionsRestrictions($matches);
+                        $entries = count(preg_split("/$splitPattern/", $response)) - 1;
```
wazelin, line 103: I'm a little confused. Shouldn't this be splitting `$value`? I mean, `$response`'s value doesn't really change in this loop.

poyuki: Defenetly, but why it works and pass tests? o_O. Will figure it out.

wazelin: That's probably because `QtiString` has a `__toString()` method ;)

poyuki: i think it shoul be something like that
```suggestion
                        $entries = count(preg_split("/$splitPattern/", $value)) - 1;
```

kilatib: It seems to me as 
$values is: `[$response->getValue()]` that's why it works

wazelin: But `$values` have nothing to do with this line ;)

poyuki: https://github.com/oat-sa/qti-sdk/blob/bfe9caaff03a65051d24d75aa3409bf56d27dffa/src/qtism/common/datatypes/QtiString.php#L95-L98
no more mistery

poyuki: i was added a test case with container but if to be honest i don't know how it can be broken cause all this pattern checks  still works with that `toString` magic of container https://github.com/oat-sa/qti-sdk/blob/c78c8185dc756d52a9f4d13e89057ec09655a7aa/src/qtism/runtime/common/Container.php#L215-L245

probably i miss something or  we can rework that loop pattern check

wazelin (approved):

PR #338, opened 2022-10-23
Abstracted Filesystem Support

serpentblade:
This PR adds support for abstracted filesystems instead of relying on Flysystem v1 exclusively.  This started as a need to support Flysystem v2+, but it was just as easy to move the Filesystem implementation in XmlDocument to interfaces in the event anybody else needs to provide their own read/write implementation.

There's an implementation of the new FilesystemInterface for Flysystemv1 and Flysystemv2+, and I've added backwards compatibility to assure XmlDocument::setFilesystem() still accepts v1 Filesystem classes directly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
 		"ext-libxml": "*",
 		"ext-mbstring": "*",
 		"ext-simplexml": "*",
-		"league/flysystem": "^1.0",
+		"league/flysystem": "^1.0|^2.0|^3.0",
 		"oat-sa/lib-beeme": "0.2.0",
 		"wp-cli/php-cli-tools": "0.10.3"
 	},
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
 		"ext-mbstring": "*",
 		"ext-simplexml": "*",
 		"league/flysystem": "^1.0|^2.0|^3.0",
+		"league/mime-type-detection": "^1.0",
 		"oat-sa/lib-beeme": "0.2.0",
 		"wp-cli/php-cli-tools": "0.10.3"
 	},
--- a/src/qtism/data/storage/xml/XmlDocument.php
+++ b/src/qtism/data/storage/xml/XmlDocument.php
@@ -122,12 +122,15 @@ class XmlDocument extends QtiDocument
         }
 
         // Support backwards compatibility of old Flysystem v1 Filesystems being passed into this class
-        if (FilesystemFactory::isFlysystemV1Installed() && $filesystem instanceof Filesystem) {
+        if ($filesystem instanceof Filesystem && FilesystemFactory::isFlysystemV1Installed()) {
             $filesystem = new FlysystemV1Filesystem($filesystem);
-        }
-
-        if (!$filesystem instanceof FilesystemInterface) {
-            throw new RuntimeException('Invalid filesystem provided.  Instance of FilesystemInterface required');
+        } elseif (!$filesystem instanceof FilesystemInterface) {
+            throw new RuntimeException(
+                sprintf(
+                    'Invalid filesystem provided.  Instance of %s required',
+                    FilesystemInterface::class,
+                )
+            );
         }
 
         $this->filesystem = $filesystem;
--- a/src/qtism/data/storage/xml/XmlDocument.php
+++ b/src/qtism/data/storage/xml/XmlDocument.php
@@ -29,8 +29,6 @@ use DOMElement;
 use DOMException;
 use Exception;
 use InvalidArgumentException;
-use League\Flysystem\Adapter\Local;
-use League\Flysystem\FileNotFoundException;
 use League\Flysystem\Filesystem;
 use LogicException;
 use qtism\common\dom\SerializableDomDocument;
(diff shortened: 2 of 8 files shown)


julien-sebire:
Hi @serpentblade . Thanks for the PR. The idea seems good to me. We're looking at it internally.

wazelin:
This looks quite good indeed. Thank you for the contribution, @serpentblade!

Could you please resolve the conflicts with the current state of https://github.com/oat-sa/qti-sdk?

src/qtism/data/storage/xml/filesystem/FlysystemV2Filesystem.php
```diff
@@ -0,0 +1,42 @@
...
+    protected Filesystem $filesystem;
+
+    public function __construct(Filesystem $filesystem)
+    {
+        $this->filesystem = $filesystem;
+    }
+
+    public static function local(string $path = '/'): self
+    {
+        return new self(new Filesystem(new LocalFilesystemAdapter($path)));
```
wazelin, line 21: Bear in mind that `XmlDocument` is a serializable object, and to keep it that way, `LocalFilesystemAdapter` instance (being composed into that object) will need to be serializable as well.

By default, though, `LocalFilesystemAdapter` will instantiate a `FinfoMimeTypeDetector` [here](https://github.com/thephpleague/flysystem/blob/3.x/src/Local/LocalFilesystemAdapter.php#L80), which will make the whole composition non-serializable due to [the `finfo` object](https://github.com/thephpleague/mime-type-detection/blob/main/src/FinfoMimeTypeDetector.php#L48).

serpentblade: It isn't clear that `XmlDocument` is intended to be `Serializable`. What's the use case of XmlDocument being serialized?  Can the filesystem be excluded during serialization and require it be re-set when unserialized? The current implementation suffers from the same potential issue as there's no implied expectation that `Filesystem` must also be serializable, and it seems a bad idea anyway to include the associated filesystem implementation in the serialized content.

wazelin: You're right on so many levels. However, it still can be kept serializable if https://github.com/thephpleague/mime-type-detection/blob/main/src/ExtensionMimeTypeDetector.php gets injected into the LocalFilesystemAdapter instance here by default.

In fact, that one should probably be sufficient for most of the use cases, as I believe the assets within a QTI package usually have their extensions in place.

src/qtism/data/storage/xml/XmlDocument.php
```diff
@@ -110,13 +113,23 @@ public function getDomDocument()
...
+            $filesystem = FilesystemFactory::local();
         }
+
+        // Support backwards compatibility of old Flysystem v1 Filesystems being passed into this class
+        if (FilesystemFactory::isFlysystemV1Installed() && $filesystem instanceof Filesystem) {
+            $filesystem = new FlysystemV1Filesystem($filesystem);
+        }
+
+        if (!$filesystem instanceof FilesystemInterface) {
+            throw new RuntimeException('Invalid filesystem provided.  Instance of FilesystemInterface required');
```
wazelin, line 130:
```suggestion
        } else {
            throw new RuntimeException(
                sprintf(
                    'Invalid filesystem provided. Instance of %s required.',
                    FilesystemInterface::class
                )
            );
```

wazelin (changes requested):

composer.json
```diff
@@ -31,7 +31,7 @@
 		"ext-libxml": "*",
 		"ext-mbstring": "*",
 		"ext-simplexml": "*",
-		"league/flysystem": "^1.0",
+		"league/flysystem": "^1.0|^2.0|^3.0",
```
wazelin, line 34: Probably last request, as the rest looks absolutely brilliant.
Now that `League\MimeTypeDetection` classes are being used explicitly in QTI-SDK's codebase, could you also include `"league/mime-type-detection": "^1.0"` here?

wazelin (changes requested):

wazelin (approved):
Superb job, thank you for adding such a great value to the project, @serpentblade!

src/qtism/data/storage/xml/filesystem/FlysystemV2Filesystem.php
```diff
@@ -0,0 +1,54 @@
...
+    public static function local(string $path = '/'): self
+    {
+        return new self(
+            new Filesystem(
+                new LocalFilesystemAdapter(
+                    $path,
+                    null,
+                    LOCK_EX,
+                    LocalFilesystemAdapter::DISALLOW_LINKS,
+                    new FallbackMimeTypeDetector(new ExtensionMimeTypeDetector()), // Serializable
```
wazelin, line 30: Two issues I failed to spot here that break the 7.4 compatibility, which we have to care about for the time being.
1. A trailing comma in the arguments' list
2. [FallbackMimeTypeDetector was added in flysystem v3 only](https://github.com/thephpleague/flysystem/commit/e8515341994dc2feb13a5723316ae521841eda0f#diff-9a576af08e4ecea9ed448f74e44c164f445a85ec447b419e8dd385d7b480685d)

wazelin (changes requested):

wazelin (approved):

wazelin:
Released as https://github.com/oat-sa/qti-sdk/releases/tag/18.1.0.

Thank you very much, @serpentblade!

PR #341, opened 2022-11-07
Feat/tr 4839/ruby make inline

kilatib:
# [TR-4839](https://oat-sa.atlassian.net/browse/TR-4839)

## Description 
When we configure Ruby in Asset tab and add it to passage , we got error `InlineCollection objects only accept to store Inline objects` 

## Development Impact
Include Ruby tags to Inline interfaces, also Include to that Group `Figure` because it should be same with `Img`

## Demo
https://user-images.githubusercontent.com/2750628/200549592-b057cb06-97f7-4d68-9801-61e5bef1984e.mov


wazelin (approved):

poyuki (approved):
LGTM

PR #342, opened 2022-11-07
Legacy: Allow ruby tag be possible to add on InlineColletions

kilatib:
# [TR-4839](https://oat-sa.atlassian.net/browse/TR-4839)

## Description 
When we configure Ruby in Asset tab and add it to passage , we got error `InlineCollection objects only accept to store Inline objects` 

## Development Impact
Include Ruby tags to Inline interfaces, also Include to that Group `Figure` because it should be same with `Img`

## Demo
https://user-images.githubusercontent.com/2750628/200549592-b057cb06-97f7-4d68-9801-61e5bef1984e.mov

### DIFF
--- a/qtism/data/content/xhtml/html5/Rt.php
+++ b/qtism/data/content/xhtml/html5/Rt.php
@@ -22,9 +22,7 @@ declare(strict_types=1);
 
 namespace qtism\data\content\xhtml\html5;
 
-use qtism\data\content\Inline;
-
-class Rt extends Html5LayoutElement implements Inline
+class Rt extends Html5LayoutElement
 {
     public const QTI_CLASS_NAME = 'rt';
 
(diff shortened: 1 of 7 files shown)


qtism/data/content/xhtml/html5/Rt.php
```diff
@@ -22,9 +22,9 @@
 
 namespace qtism\data\content\xhtml\html5;
 
-use qtism\data\content\FlowStatic;
+use qtism\data\content\Inline;
 
-class Rt extends Html5LayoutElement implements FlowStatic
+class Rt extends Html5LayoutElement implements Inline
```
wazelin, line 27: While `<figure>` and `<ruby>` must indeed be of an inline type, the `<rt>` is still a block element.
```suggestion
class Rt extends Html5LayoutElement
```

wazelin (changes requested):

wazelin (approved):

PR #343, opened 2022-11-15
feat: support php8.1 allow to serialize XmlDocument

kilatib:
# Allow XmlDocument be serializable

## Description
When we cache `XmlDocument` on php 8.1 we have error because it contain `DOMDocument` which cannot be serializable 

## Development impact 
As quick solution we use `SerializableDomDocument` and contain it inside `XmlDocument` looks like it fix the cache problem


wazelin:
Closed in favor of #344

PR #344, opened 2022-11-15
Feature: Serializable XML document

wazelin:
This PR aims to make the `XmlDocument` and its descendants serializable and stringable.

It is a backwards compatible feature, and is absolutely required since the QTI bundle has misused the XMLDocument serialization, and set the whole object to the cache [here](https://github.com/oat-sa/bundle-qti/blob/master/Factory/TestSessionAccessorFactory.php#L124), instead of explicitly calling `saveToString()` method.

Relevant commits can be reviewed as https://github.com/oat-sa/qti-sdk/pull/344/files/5185bd89945732f81cede2fad2be9a0fe78e77a7..33f8ba843cb8833b296362c24afe56d7fee88ad0


poyuki (approved):
LGTM

src/qtism/data/storage/xml/XmlDocument.php
```diff
@@ -75,6 +77,21 @@ class XmlDocument extends QtiDocument
...
+    }
+
+    public function __unserialize(array $data): void
+    {
+        $this->loadFromString($data['domDocument']);
+    }
+
+    public function __toString(): string
+    {
+        return $this->saveToString(false);
```
kilatib, line 92: Is it ok that now we could have exception on serialization if no QtiComponent was inside ?

Because that situation could be 
https://github.com/oat-sa/qti-sdk/blob/28cd6bd951161afeab9f2273e82103ca392b6390/src/qtism/data/QtiDocument.php#L60

wazelin: The only way to initialize data in `XmlDocument` is to call one of the `load` methods. That will set the component as well.

If no data was loaded, I presume there was no reason to serialize that object.

kilatib (approved):
LGTM

poyuki (approved):
LGTM

PR #348, opened 2022-11-22
feat: upgrade FlySysytem to 3 version

kilatib:
# [TR-4875](https://oat-sa.atlassian.net/browse/TR-4875)

## Description 
Upgrade the  library to use v3 of [thephpleague/flysystem](https://github.com/thephpleague/flysystem)

## Development impact
1. Adopt the lib to work with iterator when retrieve all file list.
2. User method `write` instead of `put`.
3. Use separate exceptions for File processing

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
 		"ext-libxml": "*",
 		"ext-mbstring": "*",
 		"ext-simplexml": "*",
-		"league/flysystem": "^1.0",
+		"league/flysystem": "^2.0 || ^3.0",
 		"oat-sa/lib-beeme": "0.2.0",
 		"wp-cli/php-cli-tools": "0.10.3"
 	},
--- a/src/qtism/data/storage/xml/XmlDocument.php
+++ b/src/qtism/data/storage/xml/XmlDocument.php
@@ -29,9 +29,9 @@ use DOMElement;
 use DOMException;
 use Exception;
 use InvalidArgumentException;
-use League\Flysystem\Adapter\Local;
-use League\Flysystem\FileNotFoundException;
+use League\Flysystem\Local\LocalFilesystemAdapter;
 use League\Flysystem\Filesystem;
+use League\Flysystem\UnableToReadFile;
 use LogicException;
 use qtism\common\dom\SerializableDomDocument;
 use qtism\common\utils\Url;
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -31,7 +31,7 @@
 		"ext-libxml": "*",
 		"ext-mbstring": "*",
 		"ext-simplexml": "*",
-		"league/flysystem": "^1.0",
+		"league/flysystem": "^2.0 || ^3.0",
```
poyuki, line 34: i guess it should considered as breaking change

kilatib: Not sure it possible in that repo

julien-sebire: It is possible for the main branch on which you're doing this PR, but I don't think the major bump is necessary, as the Flysystem change doesn't impact anything outside, all the changes are contained within this library.
If any doubt, proceed with major bump, but maybe check before if this impacts conflicts with the version of Flysystem optionally used in the code using this library first.

kilatib: yes, it influent for two places and it was corrected also
https://github.com/oat-sa/generis/pull/1028
https://github.com/oat-sa/lib-flysystem-filecache/pull/8

poyuki: will i be able update tao-deliver-be on that version with flysytem 1 installed?
I guess no
Screenshot 2022-11-25 at 14 37 41

it require from me additioanal adjustment (update flysystem) to be able update this library?
isn't it should be considered as breaking change?

wazelin: I agree that this is a breaking change. Not because of the dependency, though, but because the public contract actually changes.
Take `public function setFilesystem(Filesystem $filesystem = null)` method. Yes, it still accepts the same `Filesystem` instance, but the interface behind it is totally different.

wazelin (changes requested):
Compatibility with earlier versions via internal adapters done [here](https://github.com/oat-sa/qti-sdk/pull/338) looks like a good approach.

@kilatib, could you please extend your solution with it?

wazelin:
I believe https://github.com/oat-sa/qti-sdk/pull/338 is close to being finished, @kilatib.
Let's give favor to that one in order to encourage community contributions.

kilatib:
not actual anymore community cover it

PR #352, opened 2023-01-19
fix: use patched version for vulnerable league/flysystem repository

bartlomiejmarszal:
FIx Time-of-check Time-of-use (TOCTOU) Race Condition in league/flysystem

PR #358, opened 2023-05-05
Release 19.0.2

poyuki:
release for #356  and #352


kilatib (approved):
LGTM

kilatib:
@poyuki , should we add tests for it?

poyuki:
@kilatib it may be good idea, but it's private method with simple logic, I don't think we miss something scary.

PR #362, opened 2023-08-31
fix: add missing bean property

pnal:
This PR aims to fix PHP marshalling of `RandomFloat` component

### DIFF
--- a/src/qtism/data/expressions/RandomFloat.php
+++ b/src/qtism/data/expressions/RandomFloat.php
@@ -37,6 +37,7 @@ class RandomFloat extends Expression
      * The min attribute value.
      *
      * @var float
+     * @qtism-bean-property
      */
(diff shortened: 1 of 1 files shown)


src/qtism/data/expressions/RandomFloat.php
```diff
@@ -37,6 +37,7 @@ class RandomFloat extends Expression
      * The min attribute value.
      *
      * @var float
+     * @qtism-bean-property
```
gabrielfs7, line 40: I have no idea the annotation here could actually change the SDK behavior

gabrielfs7 (approved):

julien-sebire (approved):
This looks good to me. Not sure if we can test/have to fix unit tests with the attribute visible...

PR #364, opened 2023-09-06
feat: twig tags for feedback modals

bartlomiejmarszal:
Provide twig oriented feedback blocks rendering

### DIFF
--- a/src/qtism/runtime/rendering/markup/AbstractMarkupRenderingEngine.php
+++ b/src/qtism/runtime/rendering/markup/AbstractMarkupRenderingEngine.php
@@ -134,6 +134,11 @@ abstract class AbstractMarkupRenderingEngine implements Renderable
     public const CSSCLASS_ABSTRACT = 9;
 
     /**
+     * Use twig tags to generate template
+     */
+    public const TWIG_ORIENTED = 10;
+
+    /**
      * An array used to 'tag' explored Component object.
      *
      * @var array
@@ -638,6 +643,10 @@ abstract class AbstractMarkupRenderingEngine implements Renderable
             $this->includeChoiceComponent($component, $rendering);
         }
 
+        if ($this->mustTwigFeedbackComponent($component) === true) {
+            $this->twigFeedbackComponent($component, $rendering);
+        }
+
         $this->setLastRendering($rendering);
     }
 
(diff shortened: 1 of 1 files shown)


bugalot (changes requested):
From my perspective, and not change the behavior of QTI-SDK, in case you simply wish to use Twig instead of the regular PHP template oriented notation (`<?php if: ?>` and co.), is to augment the [MarkupPostRenderer::render()](https://github.com/oat-sa/qti-sdk/blob/develop/src/qtism/runtime/rendering/markup/MarkupPostRenderer.php#L224) method to make it Twig capable. It shoudln't be a big deal.

This would cover all the scenarios, using Twig, or the current implementation.

src/qtism/runtime/rendering/markup/AbstractMarkupRenderingEngine.php
```diff
@@ -638,6 +643,10 @@ protected function processNode($base = ''): void
             $this->includeChoiceComponent($component, $rendering);
         }
 
+        if ($this->mustTwigFeedbackComponent($component) === true) {
+            $this->twigFeedbackComponent($component, $rendering);
```
pribi, line 647: I propose to clear up the naming a bit. If I understand it correctly:

Does this mean if twig feedback component must be used? I propose to rename to:
if ($this->mustUseTwigFeedbackComponent($component) === true) { ... }

and set twig feedback component:
```$this->setTwigFeedbackComponent($component, $rendering);```

**Comment by bartlomiejmarszal** [comment]
> From my perspective, and not change the behavior of QTI-SDK, in case you simply wish to use Twig instead of the regular PHP template oriented notation (`<?php if: ?>` and co.), is to augment the [MarkupPostRenderer::render()](https://github.com/oat-sa/qti-sdk/blob/develop/src/qtism/runtime/rendering/markup/MarkupPostRenderer.php#L224) method to make it Twig capable. It shoudln't be a big deal.
> 
> This would cover all the scenarios, using Twig, or the current implementation.

Hey! Thank you for your comment. For me, it is not clear the reason behind using two different renderers. I saw a way how I could generate `<!-- qtism-if --!>` notation. The result of this is following:
```
<!-- qtism-if (($qtismState['PASS_CATEGORY_SCORED_RENDERING'] !== null && ($qtismState['PASS_CATEGORY_SCORED_RENDERING'] instanceof qtism\\common\\datatypes\\QtiIdentifier && $qtismState['PASS_CATEGORY_SCORED_RENDERING']->equals(new qtism\\common\\datatypes\\QtiIdentifier('not_passed')) || $qtismState['PASS_CATEGORY_SCORED_RENDERING'] instanceof qtism\\runtime\\common\\Container && $qtismState['PASS_CATEGORY_SCORED_RENDERING']->contains(new qtism\\common\\datatypes\\QtiIdentifier('not_passed'))))): -->
```
This should be replaced in `MarkupPostRenderer` with twig as a new feature. I saw this process should be done by regexp operations. But I do not understand this approach. Why do I need to complicate this like that? Why I can't just generate the correct markup when required (based on `FeedbackShowHidePolicy`)?

PR #373, opened 2023-10-04
Feat/tr 5782/dry run items

gabrielfs7:
https://oat-sa.atlassian.net/browse/TR-5782
https://oat-sa.atlassian.net/browse/RFE-544

# Goal 

- After changing Route specs like section selection or shuffle, we need to recreate the route

# Related PRs

- [ ] https://github.com/oat-sa/tao-deliver-be/pull/1212

### DIFF
--- a/src/qtism/runtime/tests/AbstractSessionManager.php
+++ b/src/qtism/runtime/tests/AbstractSessionManager.php
@@ -138,6 +138,17 @@ abstract class AbstractSessionManager
     }
 
     /**
+     * Allow to recreate route items when AssessmentTest structure is changed
+     *
+     * @param AssessmentTest $test
+     * @return Route
+     */
+    public function recreateRoute(AssessmentTest $test): Route
+    {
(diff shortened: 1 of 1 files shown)


kilatib (commented):
What the reason in that method they do nothing, why we just not use them directly? 

P.S.> Copyright (c) not updated

gabrielfs7:
> What the reason in that method they do nothing, why we just not use them directly?
> 
> P.S.> Copyright (c) not updated

Thanks @kilatib !

To answer your question, there are 2 reasons I've created new methods.

1) I needed to use the behavior of those methods outside and they are protected.
2) I did not want to change the visibility of the current methods (that are protected) as I would have to change all the places inheriting those methods
3) I am not sure that their behavior will be the only behavior needed in the future and I would not like to make the existing methods logic (internal) to be affected by this.

kilatib (approved):
LGTM

hectoras (approved):

PR #381, opened 2023-12-12
Fix TR-5932 branching rules evaluation for non-linear test parts

gabrielfs7:
https://oat-sa.atlassian.net/browse/TR-5932

Make branch rules work again between test part with nonlinear navigation

### Related PR

- https://github.com/oat-sa/tao-deliver-be/pull/1252

### DIFF
--- a/src/qtism/data/rules/BranchRuleCollection.php
+++ b/src/qtism/data/rules/BranchRuleCollection.php
@@ -31,6 +31,20 @@ use qtism\data\QtiComponentCollection;
  */
 class BranchRuleCollection extends QtiComponentCollection
 {
+
+    /** @var bool */
+    private $supportsNonLinearNavigationMode = false;
+
+    public function isAllowedForNonLinearNavigationMode(): bool
+    {
+        return $this->supportsNonLinearNavigationMode;
+    }
+
+    public function allowForNonLinearNavigationMode(): void
+    {
+        $this->supportsNonLinearNavigationMode = true;
+    }
+
     /**
      * Check if a given $value is an instance of BranchRule.
      *
--- a/src/qtism/data/rules/BranchRuleCollection.php
+++ b/src/qtism/data/rules/BranchRuleCollection.php
@@ -33,16 +33,16 @@ class BranchRuleCollection extends QtiComponentCollection
 {
 
     /** @var bool */
-    private $supportsNonLinearNavigationMode = false;
+    private $allowNonLinearNavigationMode = false;
 
-    public function isAllowedForNonLinearNavigationMode(): bool
+    public function isNonLinearNavigationModeAllowed(): bool
     {
-        return $this->supportsNonLinearNavigationMode;
+        return $this->allowNonLinearNavigationMode;
     }
 
-    public function allowForNonLinearNavigationMode(): void
+    public function allowNonLinearNavigationMode(): void
     {
-        $this->supportsNonLinearNavigationMode = true;
+        $this->allowNonLinearNavigationMode = true;
     }
 
     /**
--- a/src/qtism/runtime/tests/AssessmentTestSession.php
+++ b/src/qtism/runtime/tests/AssessmentTestSession.php
@@ -2415,7 +2415,7 @@ class AssessmentTestSession extends State
             if (
                 $ignoreBranchings === false &&
                 $numberOfBranchRules > 0 &&
-                $this->mustApplyBranchRules() || $branchRules->isNonLinearNavigationModeAllowed()
+                ($this->mustApplyBranchRules() || $branchRules->isNonLinearNavigationModeAllowed())
             ) {
                 for ($i = 0; $i < $numberOfBranchRules; $i++) {
                     $engine = new ExpressionEngine($branchRules[$i]->getExpression(), $this);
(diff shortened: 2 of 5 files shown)


src/qtism/data/rules/BranchRuleCollection.php
```diff
@@ -31,6 +31,20 @@
  */
 class BranchRuleCollection extends QtiComponentCollection
 {
+
+    /** @var bool */
+    private $supportsNonLinearNavigationMode = false;
+
+    public function isAllowedForNonLinearNavigationMode(): bool
```
wazelin, line 38: Just a suggestion
```suggestion
    public function allowsNonLinearNavigationMode(): bool
```
or perhaps
```suggestion
    public function isNonLinearNavigationModeAllowed(): bool
```

src/qtism/data/rules/BranchRuleCollection.php
```diff
@@ -31,6 +31,20 @@
...
+    {
+        return $this->supportsNonLinearNavigationMode;
+    }
+
+    public function allowForNonLinearNavigationMode(): void
```
wazelin, line 43: Another suggestion.
```suggestion
    public function allowNonLinearNavigationMode(): void
```

"Allow for" sounds more like "take into consideration" in this conext.

src/qtism/data/rules/BranchRuleCollection.php
gabrielfs7: Ok, done. Thanks!

src/qtism/data/rules/BranchRuleCollection.php
gabrielfs7: Done, thanks!

src/qtism/runtime/tests/AssessmentTestSession.php
```diff
@@ -2412,7 +2412,11 @@ protected function nextRouteItem($ignoreBranchings = false, $ignorePreConditions
             $numberOfBranchRules = $branchRules->count();
 
             // Branchings?
-            if ($ignoreBranchings === false && $numberOfBranchRules > 0 && $this->mustApplyBranchRules() === true) {
+            if (
+                $ignoreBranchings === false &&
+                $numberOfBranchRules > 0 &&
+                $this->mustApplyBranchRules() || $branchRules->isNonLinearNavigationModeAllowed()
```
wazelin, line 2418: Just a sanity check.
Did you mean what you've written, or did you mean this instead?

```
                $ignoreBranchings === false &&
                $numberOfBranchRules > 0 &&
                ($this->mustApplyBranchRules() || $branchRules->isNonLinearNavigationModeAllowed())
```

wazelin: In either case, I strongly advise using parenthesis (even if the order of logical operations is correct) to emphasize the intended logic.

gabrielfs7: I agree with you, let me improve that

wazelin: Oh. So, it _was_ a wrong condition after all?
I mean, what you've applied now is totally different from the original condition where, should `$branchRules->isNonLinearNavigationModeAllowed()` be `true` the rest of the condition didn't matter.

wazelin: Now the question is: "why didn't the test cases fail with the original code"?

gabrielfs7: The test cases did not fail before, cause the test part navigation mode for those cases were always linear. I've extended the tests with nonlinear support. So now, if the branch rules is coming from the tests part (las one to be applied), we ignore the linear mode restriction.

wazelin: No-no. I mean why were they working when the condition was
```
                $ignoreBranchings === false &&
                $numberOfBranchRules > 0 &&
                $this->mustApplyBranchRules() || $branchRules->isNonLinearNavigationModeAllowed()
```
?

It's very different from the current one.

gabrielfs7: We probably do not have proper coverage for those conditions

wazelin: That's fine. Just once again want to be 100%-sure your changes are intentional.
Quick reference of the changes between the two conditions we saw in this PR before and after my comment on this line: https://3v4l.org/G5cLN

gabrielfs7: Yes, it is intentional, before the `||` operator outside of the parentheses would invalidate all the time all the other rules (which was passing cause of low test coverage), but it only makes sense to pass if other conditions are met too :) 

Thanks for the heads up

wazelin (approved):

src/qtism/runtime/tests/AssessmentTestSession.php
wazelin: OK then.
I just didn't realize that from your initial response to my comment.

All clear now.

pnal (approved):

src/qtism/runtime/tests/AssessmentTestSession.php
gabrielfs7: FYI, @ ignoreBranchings and other parameters for this very method are never passed. They were introduced on 2020, but never used. I will not change them now to avoid inflate the PR, but that is also one of the reasons that condition is not caught on tests

PR #397, opened 2024-06-06
fix: not validate maxWords for XHTML extendTextInteractions

kilatib:
# Skip validation for ExtendTextINteraction [TR-6169](https://oat-sa.atlassian.net/browse/TR-6169)
https://oat-sa.atlassian.net/browse/TR-6169

## Description 
Skip validation for size on XHTML text 

## Related PRs
- https://github.com/oat-sa/tao-deliver-be/pull/1346

[TR-6169]: https://oat-sa.atlassian.net/browse/TR-6169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/qtism/data/content/interactions/ExtendedTextInteraction.php
+++ b/src/qtism/data/content/interactions/ExtendedTextInteraction.php
@@ -162,6 +162,9 @@ class ExtendedTextInteraction extends BlockInteraction implements StringInteract
      */
     private $format = TextFormat::PLAIN;
 
+    // This option disable also validation for patternMask
+    private $isDisabledMaxWordValidation = false;
+
     /**
      * Create a new ExtendedTextInteraction object.
      *
@@ -313,7 +316,7 @@ class ExtendedTextInteraction extends BlockInteraction implements StringInteract
      */
     public function getPatternMask(): string
     {
-        return $this->patternMask;
+        return !$this->isDisabledMaxWordValidation() ? $this->patternMask : '';
     }
 
     /**
@@ -323,7 +326,7 @@ class ExtendedTextInteraction extends BlockInteraction implements StringInteract
      */
     public function hasPatternMask(): bool
     {
-        return $this->getPatternMask() !== '';
+        return !$this->isDisabledMaxWordValidation() && $this->getPatternMask() !== '';
     }
 
     /**
@@ -389,7 +392,7 @@ class ExtendedTextInteraction extends BlockInteraction implements StringInteract
      */
     public function getMaxStrings(): int
     {
-        return $this->maxStrings;
+        return !$this->isDisabledMaxWordValidation() ? $this->maxStrings : -1;
     }
 
     /**
@@ -399,7 +402,7 @@ class ExtendedTextInteraction extends BlockInteraction implements StringInteract
      */
     public function hasMaxStrings(): bool
     {
-        return $this->getMaxStrings() !== -1;
+        return !$this->isDisabledMaxWordValidation() && $this->getMaxStrings() !== -1;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/qtism/data/content/interactions/ExtendedTextInteraction.php
```diff
@@ -313,7 +313,7 @@ public function setPatternMask($patternMask): void
      */
     public function getPatternMask(): string
     {
-        return $this->patternMask;
+        return $this->getFormat() !== TextFormat::XHTML ? $this->patternMask : '';
```
wazelin, line 316: That's not exactly what I meant. We shouldn't skip any validations by default.
QTI SDK should offer a way to optionally disable certain validations.

wazelin (changes requested):

src/qtism/data/content/interactions/ExtendedTextInteraction.php
```diff
@@ -162,6 +162,9 @@ class ExtendedTextInteraction extends BlockInteraction implements StringInteract
      */
     private $format = TextFormat::PLAIN;
 
+    // This option disable also validation for patternMask
+    private $isDisabledMaxWordValidation = false;
```
poyuki, line 166: could you please clarify a bit where  and how the property is used? can't find it in this PR and related

kilatib: sorry, yes it is not used

src/qtism/data/content/interactions/ExtendedTextInteraction.php
```diff
@@ -505,7 +507,13 @@ public function getResponseValidityConstraint(): ResponseValidityConstraint
...
             $this->getMinStrings(),
             ($this->hasMaxStrings() === false) ? 0 : $this->getMaxStrings(),
-            $this->getPatternMask()
+            $this->getPatternMask(),
+            [
+                'qtiClassName' => $this->getQtiClassName(),
+                'options' => [
+                    'format' => $this->getFormat(),
+                ],
+            ]
```
wazelin, line 516: What if instead of loosely-typed `extraData` we provisioned a reference to the whole instance of the current `Interaction` here?
```suggestion
            $this
```

Instead of `extraData` property on the constraint level, you can then have `private ?Interaction = null`.

Do you foresee any problems with that? If so, or if you don't think it makes any sense, ignore this comment.

kilatib: Yes, it was my first idea, but we have several problems with it on serialized and unserialized actions for the final session. It is not super practice, but it is better to keep here simple data.

src/qtism/data/content/interactions/ExtendedTextInteraction.php
```diff
@@ -473,7 +475,7 @@ public function hasExpectedLines(): bool
     /**
      * Set the format of the text entered by the candidate.
      *
-     * @param int $format A value from the TextFormat enumeration.
+     * @param  int $format A value from the TextFormat enumeration.
```
wazelin, line 478: I don't think we need to align the docblock elements like that.
I'd rather have all of them consistent and with a single space as a delimiter.

wazelin (changes requested):
Just a tiny nitpick, and we're good-to-go!

poyuki (approved):
LGTM nicely done. Approving it as doc block  changes are requested to be rolled back.

src/qtism/data/state/ResponseValidityConstraint.php
```diff
@@ -66,24 +66,28 @@ class ResponseValidityConstraint extends QtiComponent
...
     /**
      * The collection of nested AssociationValidityConstraints objects.
+     *
+     * @var AssociationValidityConstraintCollection
      */
     private $associationValidityConstraints;
 
     /**
-     *  Provide additional information about the ResponseValidityConstraint.
+     * Metadata defined by @see \qtism\data\content\interactions\Interaction instantiating this ResponseValidityConstraint
```
kilatib, line 75: Thank you

PR #399, opened 2024-06-28
feat: modyfy qti identifier regex

bartlomiejmarszal:
This will fix remote publication process that is failing on new identifier format

### DIFF
--- a/src/qtism/common/utils/Format.php
+++ b/src/qtism/common/utils/Format.php
@@ -97,7 +97,7 @@ class Format
     public static function isIdentifier(string $string, bool $strict = true): bool
     {
         if (!$strict) {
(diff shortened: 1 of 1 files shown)


julien-sebire (changes requested):
⚠️ This must not be done as it would break compatibility with QTI. Identifiers can’t begin with a digit.

bugalot (approved):
Error, I didn't want to approve :upside_down_face:

bugalot (changes requested):
Hey guys :) ! We can't do this, unfortunately. This rule is expressed by the QTI XSD Schema. Letting this happen means that we accept items not matching the official IMS validator.

PR #402, opened 2024-07-11
fix: AssessmentItem::setIdentifier uses isNormalizedString

bartlomiejmarszal:
This allow to set identifier that is validated against normalized string


bugalot (changes requested):
Unit tests required including a sample that was failing before required ;)

bugalot:
We would like to also see the backport coming, otherwise we will for sure forget. I would strongly suggest to merge this in `develop`, and simply release in order not to miss this in next release.

bugalot:
OK, now I understand, it's the fix on the legacy version. However, please update the unit tests as well, so that it goes along that version too.

tikhanovichA (approved):

PR #410, opened 2025-01-20
fix: null-safe dereferencing of record-typed variables fields via the `<fieldValue>` operand

wazelin:
# [COECDP25-952](https://oat-sa.atlassian.net/browse/COECDP25-952)

This PR makes `FieldValue` processor less strict, and resolve into a `null` value in case the referenced variable itself is a `null`.

e.g. the following rule will produce a `null` instead of failing with an exception if the `RESPONSE` variable is not set at all.

```xml
<fieldValue fieldIdentifier="result">
  <variable identifier="RESPONSE"/>
</fieldValue>
```

In other words, instead of treating that expression as `RESPONSE.result`, we will treat it as `RESPONSE?.result`.

According to https://www.imsglobal.org/sites/default/files/spec/qti/v3/info/index.html#Data_FieldValue, the 
> The field value operator takes a sub-expression with a record container value. The result is the value of the field with the specified field-identifier. If there is no field with that identifier then the result of the operator is NULL.

Arguably, there is no field with any identifier if the whole variable cannot be dereferenced.

[COECDP25-952]: https://oat-sa.atlassian.net/browse/COECDP25-952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Before the fix

https://github.com/user-attachments/assets/37dc23d3-849a-426a-b902-aafe1fc955de

## After the fix

https://github.com/user-attachments/assets/05a964b6-a76e-46b4-8c88-e01668a5cd92

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -6,16 +6,23 @@ on:
   pull_request:
     branches: [ develop ]
 
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
 
-    runs-on: ${{ matrix.operating-system }}
+    runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
       matrix:
-        operating-system: [ ubuntu-latest ]
-        php-versions: [ '7.4', '8.0', '8.1', '8.2', '8.3']
+        php-version: [ '7.4', '8.0', '8.1', '8.2', '8.3' ]
+        coverage: [ false ]
+        include:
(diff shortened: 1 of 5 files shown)


kilatib (approved):
- [x] Commits are following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] Changelog is updated according to changes (if applicable)
- [ ] Documentation is updated according to changes (if applicable)

PR #412, opened 2025-02-05
chore: fixing syntax throwing an error

uncleempty:
It is there to fix a small problem, introduced in one of the previous PRs.

As per the customer report, it shows syntax errors when trying to run the command from the SDK readme:
```sh
./vendor/bin/qtisdk render -df --source /Users/*/Downloads/i6422d9afb42b32279aab31294164637a8/qti.xml --flavour goldilocks > ./qti.html
PHP Parse error:  syntax error, unexpected token "{" in /Users/*/Downloads/test/taoApi/vendor/qtism/qtism/src/qtism/cli/Render.php on line 283
```
Which is pretty logical, since variable interpolation won't work outside of "string context".

### DIFF
--- a/src/qtism/cli/Render.php
+++ b/src/qtism/cli/Render.php
@@ -280,14 +280,14 @@ class Render extends Cli
                 $body = substr($body, 0, strlen('</div>') * -1);
                 $body = "<body {$body}</body>{$nl}";
             } else {
-                $body = $xml->saveXml($xml->documentElement) . {$nl};
+                $body = $xml->saveXml($xml->documentElement) . $nl;
             }
 
             if ($arguments['document'] === true) {
                 $footer = "</html>\n";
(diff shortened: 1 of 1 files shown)


wazelin (approved):

wazelin:
Ooh. This is pretty bad.
Thank you for the fix, @uncleempty!

Luckily, this issue concerns only the CLI part of the library. Merging and releasing should be sufficient. No need to deploy this anywhere.

bugalot:
Thanks for this! You might also want to consider upgrading wp-cli as in the following PR (that I never merged)
CC @wazelin / @uncleempty https://github.com/oat-sa/qti-sdk/pull/408

bugalot:
as well as the usage of relative paths in that exact command, which would be nice :)

PR #422, opened 2025-06-18
fix: more even shuffling algorithm

pnal:


KirylHatalski (approved):

src/qtism/runtime/rendering/markup/xhtml/Utils.php
```diff
@@ -71,53 +71,85 @@ public static function shuffle(DOMNode $node, ShufflableCollection $shufflables)
...
+            }
 
-                unset($placeholder1);
-                unset($placeholder2);
+            unset($placeholder1);
+            unset($placeholder2);
+        }
+    }
+
+    public static function getSwappingMapByValues(array $shufflableIndexes, array $shuffledIndexes): array
```
Karol-Stelmaczonek, line 122: Can you add some comment with description and example what the function does?

